PR #12, opened 2017-12-07
Allow hostname changes from status page

dbuxo:


dbuxo:
/etc/hostname is currently readonly, these patches add a custom hostname config file on the SDcard

EliasKotlyar:
Thanks you for your contribution!

Issue #63, opened 2018-02-02
RTSP autostart


hondaj:
Thank you for your hard work! I have a question is it possible to „autostart” RTSP or any other setting when reboot device?

kszere:
You can it make adding last line to `run.sh` file. 
```
MicroSD
├── bin
├── config
├── run.sh
├── scripts
├── www
└── run.sh
```
## I use this
#### For play welcome sound after start camera.
`/system/sdcard/bin/ossplay -g 100 /system/sdcard/sounds/startup.wav &`
`-g` param is for volume settings, 100 is 100%, but you can set 500 (500%)

#### For autostart stream RTSP
`/system/sdcard/bin/busybox nohup /system/sdcard/bin/v4l2rtspserver-master -fMJPG -F 10 &>/dev/null &`

shinebar1001:
Just added a small pull-request to make this configurable

hondaj:
Thanks for the hint, it works perfectly! 
After many hours of testing and work around with Synology, I'm also able now to connect the Dafang to Surveillance 8.1, streaming/record well, no audio in MJPEG, no cam controll but it's another topic. Thank you once again!

PR #65, opened 2018-02-02
Added configurable autostart-options for rtsp and auto-night-mode & auto night-vision toggle for rtsp-server

shinebar1001:
Again a very basic addition: some config-files and  added some lines of code for configuring the autostart of rtsp-h264 || mjpeg and the auto-night-detection.
No fancy stuff, but working
#63 #37


kszere:
Good idea, but i think that better add script to `/system/sdcard/scripts/` with for example run rtsp command.
Next step add symlink to `/system/sdcard/config/autostart/`
Next step in run.sh, find all symlinks in up folder and run all. If not need, delete symlink.
Next step in action.cgi or api.cgi add run command from `/system/sdcard/scripts/`


It is my propose. :)

shinebar1001:
That sounds like a very good idea  to get the things a little bit modular.  Will see, when I can spend some time on that topic. Currently I've nearly finished a nightvision-toggle for the rtsp-server and an updated current image with the avconv from the thread you mentioned. :-)

In my opinion it's necessary to get the basic functions running as soon as possible to get a broader userbase and more contributors - so it's sometimes what architects call "form follows function"

shinebar1001:
Okay, here we go - I made use of the "Manage Scripts" Tab, and wrote some scripts for starting / stopping the rtsp-server and so on. Symlinks were not possible, because the sdcard is FAT, so I had to take a workaround.

Additionally I included the avconv from the forum and updated the currentpic.cgi to show a picture whenever possible. 

image

PR #85, opened 2018-02-10
 OSD configurable with webinterface

shinebar1001:
Works quite well with one problem - probably you've got an idea?
I don't have a clue how to use a variable with a blank when starting v42lrtspserver, tried all combinations with quotation marks, double, single whatever. The OSD-Text will end after the first blank:

```
[root@Pucki-Cam:tmp]# cat /system/sdcard/config/osd
OSD='-D "DafangHacks %H:%M:%S %d.%m.%Y" -d UP'
[root@Pucki-Cam:tmp]# /system/sdcard/bin/v4l2rtspserver-master -fMJPG $OSD &
```
(skipped the notices)
the RTSP-Server ist running, OSD is **"Dafang Hacks** (including the quotation marks)

The command is alright anyway
```
[root@Pucki-Cam:tmp]# ps w | grep v4l
 1455 root      129m R    /system/sdcard/bin/v4l2rtspserver-master -fMJPG -D "DafangHacks %H:%M:%S %d.%m.%Y" -d UP
```

if I start this from the command line everything is fine and the OSD works as designed.

Putting the same string in a variable like
```
[root@Pucki-Cam:tmp]# TEST='/system/sdcard/bin/v4l2rtspserver-master -fMJPG -D "DafangHacks %H:%M:%S %d.%m.%Y" -d UP'
[root@Pucki-Cam:tmp]# echo $TEST
/system/sdcard/bin/v4l2rtspserver-master -fMJPG -D "DafangHacks %H:%M:%S %d.%m.%Y" -d UP
[root@Pucki-Cam:tmp]# $TEST
```

Im sure the answer cannot be that hard, but I'm lost.

So every blank is replaced by a underscore - thats the only problem until now. 

best, 
danny

image

### DIFF
--- a/firmware_mod/config/osd
+++ b/firmware_mod/config/osd
@@ -1,0 +1,1 @@
+OSD="-D DafangHacks_%H:%M:%S_%d.%m.%Y  -d UP"
--- a/firmware_mod/controlscripts/rtsp-h264
+++ b/firmware_mod/controlscripts/rtsp-h264
@@ -3,6 +3,10 @@ PIDFILE="/run/v4l2rtspserver-master-h264.pid"
 NIGHTFILE="/run/v4l2rtspserver-master-h264.night"
 export LD_LIBRARY_PATH='/thirdlib:/system/lib'
 
+if [ -f /system/sdcard/config/osd ]; then
+source /system/sdcard/config/osd  2>/dev/null
+fi
+
 status()
 {
   pid="$(cat "$PIDFILE" 2>/dev/null)"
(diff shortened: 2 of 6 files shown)


shinebar1001:
Added an averaging for the nightvision
image

shinebar1001:
Added cURL with libressl to up / download files (targeting a cyclic upload)
cleaned up network.cgi a little bit - doesn't show the configure options anymore but now shows the SSID in use (just had to update the path)

PR #89, opened 2018-02-11
Fixed even more files & added mqtt-control

shinebar1001:
@EliasKotlyar had to dos2unix some more files... 
Don't really know how i scrambled the files but will definitely update my QA-process. Sorry.


shinebar1001:
Added mqtt-control. You have to edit the config file before you are able to start the script. I think there is no need to make this configurable via web, because someone who uses mqtt seems to have enough knowledge to edit a file and rename it.

PR #92, opened 2018-02-13
added a mqtt-status + jq

shinebar1001:


pplucky:
One question on mqtt status: will a request to the camera needs to be sent, for it to publish its status, or will the camera send it on itself periodically? Second option would be preferable, if possible.

shinebar1001:
At the moment you need to send status_text to the camera and you'll get that reply. A periodical status is no problem and already on my personal roadmap.
Which interval would you suggest? More like every 5 minutes or more like once a minute?

pplucky:
What I'd would be to define a parameter in the mqtt config file and do it based on the value assigned to it. Maybe define a minimum of 30sec under which it will not go, regardless of the config and allow anything above as per user setting.

lolouk44:
I would second @pplucky let people choose but make it so that no less than 30 sec

shinebar1001:
hey guys, the feature is added in e981159eade2b0e27ab420b498d5f37062daab5c - you'll need to update your mqtt-config, please see /system/sdcard/config/mqtt.dist and copy the last two Parameters.

PR #103, opened 2018-02-18
Add reboot to interface live

gysmo38:
Try my first pull request....

Not sure it is the good way.


shinebar1001:
that's the right way 👍

shinebar1001:
@gysmo38 there are several links on cam_entree - can you make these relative?

gysmo38:
Ok I think you correct it because it look good now.

I will be more careful next time.

PR #108, opened 2018-02-19
Update mqtt.dist

jmtatsch:
set STATUSINTERVAL variable, without it there will be 3-4 status updates per second driving the load to : 4.24, 4.29, 3.27"


shinebar1001:
Thank you! Forgot to update the mqtt.dist...

PR #110, opened 2018-02-19
Favicon for status webpage + MQTT IR-Cut status fix

pplucky:
Status webpage is prepared to show a favicon, but none was provided + MQTT status for IR-Cut was wrong


shinebar1001:
We have to look in the IR_Cut Thing again. In my opinion IR Cut is on when there is daylight to have good Colors and off at night to view light in the IR-spectrum as well. (https://en.wikipedia.org/wiki/Infrared_cut-off_filter) - it seems to be mixed up again, or am I on the wrong path?

PR #140, opened 2018-03-03
Add html configuration for OSD, need new vl2rtspserver-master and set…

nik0:
Add html configuration for OSD, need new vl2rtspserver-master and setconf
Modify run.sh to set OSD conf at startup


EliasKotlyar:
Hello nik0,

Thanks for your support. I have merged it!

PR #152, opened 2018-03-09
Remove extra backtick

Cabalist:
I was reviewing the code after purchasing one of these cameras.  I noticed that there *may* be an extra backtick here.  I can't currently run this code (camera is still on the way) but thought I should point it out.


shinebar1001:
Thanks!

PR #163, opened 2018-03-16
fix misleading ir cut state

jmtatsch:
ir cut/filter on is needed with daylight, ir cut off at night when we use the ir leds to illuminate the scene.


shinebar1001:
@jmtatsch at the moment github complains conflicts that must be resolved. I don't see the conflicts... This is why I didn't merge your PR like the two other ones.

jmtatsch:
@shinebar1001 should be ok now. I accidentally pushed another feature and it got added to the same pr.

PR #177, opened 2018-03-23
Enhance live ui with responsive switches for many functions

jmtatsch:


wojtulab:
@jmtatsch is it possible to add MQTT button to send a custom command to MQTT broker?

jmtatsch:
@wojtulab what would you use that for?

wojtulab:
@jmtatsch I have custom MQTT sounds actions defined in Dafang.
I would like to play it from live.html using mqtt buttons.

PR #184, opened 2018-03-25
add some config files

theyort:
Change to get a private NTP server and the associate config file.
A start of change to choose the video size
Remove of the start of the RTSP server in the run.sh, because in case of a select in the scripts page, it appears that the process has been already started
TheYort


EliasKotlyar:
Hello theyort,

Many thanks for your contribution! I have merged it.

Greatings
Elias

lenisko:
@theyort ` Unsupported command 'setvideosize' `

PR #192, opened 2018-03-30
Fix mqtt status by removing doublequoting

jmtatsch:
thanks @tthk


jmtatsch:
We can as well go ahead and merge this part already as @tthk's pr now has conflicts anyway.

PR #199, opened 2018-03-31
Fix syntax errors to allow MQTT to work.  Use pipe not FIFO.

tthk:
First time doing pull request, let me know if this is the correct way.

Fixed syntax errors in the two files to allow MQTT to work as intended.  Added MQTT publish to inform status after MQTT control requests.


jongdesteven:
You are breaking the functionality of the response of "/topic/blue_led/set on|off"
this should be as it was; response per setting on "topic/blue_led" with on or off. Not the status report.

tthk:
@jongdesteven 
> You are breaking the functionality of the response of "/topic/blue_led/set on|off"
this should be as it was; response per setting on "topic/blue_led" with on or off. Not the status report.

We can certainly put back the responses to setting the LEDs if these are expected (I'm new to MQTT protocols and etiquettes).  I thought publishing only the status, which includes the true states of LEDs would be cleaner.  If turning on blue LED didn't turn off yellow LED in the future, we'd wouldn't have to change scripts/mqtt-control.sh if it just responded with the true LEDs' states.

Here is my HomeAssistant configuration.yaml entries for the LEDs, which works well based on only the complete JSON status response:
```
  - platform: mqtt
    name: "Dafang Living Room Blue LED"
    state_topic: "LivingRoom/DaFang/status"
    value_template: "{% if value_json.LEDs.Blue == 'on' %}on{% else %}off{% endif %}"
    command_topic: "LivingRoom/DaFang/Blue_LED/set"
    payload_on: "on"
    payload_off: "off"
    optimistic: false
  - platform: mqtt
    name: "Dafang Living Room Yellow LED"
    state_topic: "LivingRoom/DaFang/status"
    value_template: "{% if value_json.LEDs.Yellow == 'on' %}on{% else %}off{% endif %}"
    command_topic: "LivingRoom/DaFang/Yellow_LED/set"
    payload_on: "on"
    payload_off: "off"
    optimistic: false
```

jongdesteven:
Checking the GPIO status before sending back the status might improve it. But i'd rather would avoid the need for a value_template.

tthk:
@jongdesteven 
>Checking the GPIO status before sending back the status might improve it. But i'd rather would avoid the need for a value_template.

Isn't Dafang-Hacks' status JSON MQTT message that's currently in use already forcing the value_template in HomeAssistant?  How would the HomeAssistant configuration.yaml entry for the LEDs look if it needs to update based on the JSON status message AND the specific MQTT responses to LED/set messages?  I'm new to HomeAssistant as well.

jongdesteven:
@tthk 
I've started to think about it for a bit and I still think that ideally MQTT should be a "Push" device. pushing out updates to things that have changed, rather than sending a global status. BUT the Dafang has multiple interface and scripts that change settings; maybe your change to sending out a status when changing something.

I think you can even simplify your value_template to:
`value_template: "{{ value_json[LEDs][Yellow] }}"`
because the response is already 'on' OR 'off'

tthk:
Ageed that the JSON full status seems to be the best way forward now.

I tried your elegant suggestion for the HomeAssistant configuration.yaml file entries for the LEDs:
```
value_template: "{{ value_json[LEDs][Yellow] }}"
```
but it didn't work.  This seems to work, though:
```
value_template: '{{ value_json["LEDs"]["Yellow"] }}'
````
Here is my configuration.yaml excerpt now, including entries to control and track RTSP servers, working well with current proposed git submit:
```
  - platform: mqtt
    name: "DaFang Living Room RTSP H264"
    state_topic: "LivingRoom/DaFang/status"
    value_template: "{% if value_json[\"RTSP-Server\"] == 'running' %}h264_start{% else %}rtsp_stop{% endif %}"
    command_topic: "LivingRoom/DaFang/set"
    payload_on: "h264_start"
    payload_off: "rtsp_stop"
    optimistic: false
  - platform: mqtt
    name: "DaFang Living Room RTSP MJPEG"
    state_topic: "LivingRoom/DaFang/status"
    value_template: "{% if value_json[\"RTSP-Server\"] == 'running' %}mjpeg_start{% else %}rtsp_stop{% endif %}"
    command_topic: "LivingRoom/DaFang/set"
    payload_on: "mjpeg_start"
    payload_off: "rtsp_stop"
    optimistic: false
  - platform: mqtt
    name: "Dafang Living Room Blue LED"
    state_topic: "LivingRoom/DaFang/status"
    value_template: '{{ value_json["LEDs"]["Blue"] }}'
    command_topic: "LivingRoom/DaFang/Blue_LED/set"
    payload_on: "on"
    payload_off: "off"
    optimistic: false
  - platform: mqtt
    name: "Dafang Living Room Yellow LED"
    state_topic: "LivingRoom/DaFang/status"
    value_template: '{{ value_json["LEDs"]["Yellow"] }}'
    command_topic: "LivingRoom/DaFang/Yellow_LED/set"
    payload_on: "on"
    payload_off: "off"
    optimistic: false
```

**Comment by tthk** [comment]
Resolved conflicts and tested.  Everything works on my end.

**Comment by jmtatsch** [comment]
@tthk Thank you so much for your contribution. Like @jongdesteven I would also prefer to keep the status update as small as possible. We should not forget we are on very resource limited device and a full mqtt-status.sh does a shitload of potentially expensive iocalls. That could easily cause disruptions to avstreams etc.

Update: I actually measured the execution time of the scripts, its 70 ms, not as bad as i have expected. 

```
time ./mqtt-status.sh
{"Uptime":" 10:43:44 up 1 day, 21 min,  0 users,  load average: 3.40, 3.49, 3.62", "RTSP-Server":"running", "IR-Cut":"off", "Wifi":{"SSID":"MyNet", "Bitrate":"72.2 Mb/s", "SignalLevel":"63%", "Linkquality":"80%", "NoiseLevel":"0%"}, "LEDs":{"Blue":"off", "Yellow":"off", "Infrared":"off"}}
real	0m 0.26s
user	0m 0.02s
sys	0m 0.00s

time ./simple_update.sh 
real	0m 0.19s
user	0m 0.03s
sys	0m 0.00s
```

**Comment by tthk** [comment]
As stated, I'm very new to mqtt and its best practices, and thus hesitating to make big changes.  The mqtt-status.sh triggering a publish once on rare activites like LED setting, camera movements, etc.., then additionally once every 30s doesn't seem too bad at a cost of 70ms, as you noted.

I do get your wishes to be as lean as  possible, on the camera side as well as mqtt chatter.  Is there a best practice guideline for mqtt chatter I can study?

Any changes we want to make in this area is probably best done soon before folks depend to much on the existing mqtt formats.

**Comment by jmtatsch** [comment]
@tthk I am also new to mqtt so I cannot offer any best practice guidelines 😃
Your implementation is probably also an improvement by requesting the changed states instead of just assuming they were changed correctly.
Let's worry about the performance when we really feel a performance impact.

Regarding your comment for stable interfaces, I really hate parsing nested jsons with templates (https://github.com/home-assistant/home-assistant/pull/13177) although that seems to be the route the HA authors want to go. We still should attempt to make the configuration as simple as possible.

I would thus propose :

- not using a complicated to parse (with templates) nested json at all and split it all up into lean individual entities (and eventually define then via autoconf) or at least flattening the json already on the camera side
- it seems to be best practice to use the entity name without status at the end for the state topic (https://www.hivemq.com/blog/mqtt-essentials-part-5-mqtt-topics-best-practices , HA uses this inconsistently) i.e. LivingRoom/DaFang/Blue_LED instead of LivingRoom/DaFang/Blue_LED/status
- it seems to be best practice to use lowercase &  / names only (https://www.hivemq.com/blog/mqtt-essentials-part-5-mqtt-topics-best-practices) i.e. livingroom/dafang/leds/blue/status & livingroom/dafang/leds/blue/set
- using home assistant defaults (ON/OFF) where possible so that we don't need extra qualifying for   payload_on & payload_off (https://www.home-assistant.io/components/switch.mqtt/). Warning: This is actually wrong in the documentation, the default payload is not true/false but ON/OFF
- if we always use state topics, we don't need an additional optimistic mode flag as it defaults to false if a state topic defined.

What do you think?

**Comment by shinebar1001** [comment]
In my opinion there is not one truth but there are many. The "all-in-one" mqtt-status is consuming many ressources but is doing a great job when you use for example things like node red - you'll get a nice object which can be used in many ways. But running this every thirty seconds is quite heavy. (I hate parsing nested JSON as well, but node red or jq does a real good job on this :))

Getting a "real" feedback to things like switching LED on or off is a nice idea, but as there is only an echo to something in /proc you can assume that this works and immediately.

So I would recommend to keep the "all in one status" as an option and default to the independent entities.

**Comment by tthk** [comment]
Thanks for the MQTT Best Practice doc pointer, @jmtatsch.

How about for important items that users may change, we respond at time of change, and JSON status for all of the miscellaneous pieces of info that people would seldom need to know immediately, but may subscribe for en-bulk interval updates, then parse out themselves.

```
{"Uptime":" 10:43:44 up 1 day, 21 min,  0 users,  load average: 3.40, 3.49, 3.62", "RTSP-Server":"running", "IR-Cut":"off", "Wifi":{"SSID":"MyNet", "Bitrate":"72.2 Mb/s", "SignalLevel":"63%", "Linkquality":"80%", "NoiseLevel":"0%"}, "LEDs":{"Blue":"off", "Yellow":"off", "Infrared":"off"}}
```
So for the current items camera reports, LED, IR-Cut, would each have its own state updates, while Uptime, RTSP-Server status, Wifi status would fall in JSON interval updates.

jmtatsch:
@tthk agreed, i'll whip something up and make a pr the next days. I'll also add a common function library that should reduce duplicate code in all the cgi pages & mqtt status & control.

Does anyone have more opinion on interface, naming etc? Especially the ON/OFF vs on/off as common internal state for most services, leds, actors.

jongdesteven:
@jmtatsch Would it be a good idea, if making a central lib, to make that central lib also in charge of keeping track of the settings. With that I mean; if you are centrally taking care of changes to the system, reporting it back to mqtt control / status would be only done on changes?

jmtatsch:
@jongdesteven that could eventually make sense. However we should at least  have some asynchronous mqtt publisher, otherwise the whole thing will get blocked if the mqtt broker is down?

lolouk44:
Nice work :)

Issue #186, opened 2018-03-26
rtsp-h264 NOK autostart


wojtulab:
Hello,
I noticed that rtsp-h264 can't start in autostart.
After manual STOP&START it's wokring normally.

After boot:
![image](https://user-images.githubusercontent.com/14801148/37912784-879bfc2a-3113-11e8-8e77-fe78fc3a6a51.png)
![image](https://user-images.githubusercontent.com/14801148/37912796-912028b6-3113-11e8-86fa-858e4afed048.png)

After Start & Stop:
![image](https://user-images.githubusercontent.com/14801148/37912876-c1ef1dda-3113-11e8-9fe7-5ef77c67ef53.png)
![image](https://user-images.githubusercontent.com/14801148/37912887-c8350b6e-3113-11e8-9dbd-a33ff68feb8a.png)

as you can see in both examples the process are running but the status is wrong

tthk:
The v4l2rtspserver-master is started in run.sh:
```
#Start
/system/sdcard/bin/busybox nohup /system/sdcard/bin/v4l2rtspserver-master -S &>/dev/null &
```
If these lines are changed to:
```
#Start
/system/sdcard/controlscripts/rtsp-h264 start
```
then this problem will go away.  Additionally, any customizations to the rtsp-h264 command can be localized in one place: controlscripts/rtsp-h264 

I'll open a pull request for this fix.  Let me know if this is the proper way to contribute fixes.

tthk:
Took some tries but I think I created the pull request for just this fix properly.

jmtatsch:
should be fixed thanks to @tthk, please re-open if not

PR #170, opened 2018-03-18
authentication for rtsp-streams added

herrxyz:
as written in my issue, here is the commit with authentication (commented out original line, you could clean it up or rename var and filename if you want)

PR #203, opened 2018-04-01
Cleaner autostart for rtsp. Fix for Issue #186

tthk:
When autostarting rtsp server, instead of doing so explicitly in run.sh, now call the controlscripts/rtsp scripts.  This allows localization of the start and stop of the service.  Also fixes problem reported in Issue #186.

### DIFF
--- a/firmware_mod/config/rtspserver.dist
+++ b/firmware_mod/config/rtspserver.dist
@@ -1,0 +1,17 @@
+#######################################################################
+# Edit this file and move it to /system/sdcard/config/rtspserver.conf #
+# if you need to add options, otherwise defaults will be used         #
+#######################################################################
+
+# Configure RTSP Servers
+# Use /system/sdcard/bin/v4l2rtspserver-master -h to see available options
+
+# H264 RTSP server options
+# Examples:
+# RTSPH264OPTS="-S -W960 -H540"
+RTSPH264OPTS=""
+
+# MJPEG RTSP server options
+# Examples:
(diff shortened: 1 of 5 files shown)


jmtatsch:
Also a centralized rtsp.conf for things like resolution and others would be great...

tthk:
Just noticed config/autostart.  Please hold off on this pull request until I code a solution to start rtsp server from the config/autostart rather than run.sh.

tthk:
@jmtatsch commented 32 minutes ago 
> Also a centralized rtsp.conf for things like resolution and others would be great...

Agreed.  Currently, settings for H264 are set in controlscripts/rtsp-h264, and MJPEG in controlscripts/rtsp-mjpeg, directly on the command line of the call to the rtsp server.

Migrating these settings to config/rtsp.conf seems like a good idea?  I'll look into doing this if there are no objections.

tthk:
Cleaned up `controlscripts/` files involved in this fix to not show irrelevant errors.
Add `config/rtspserver.dist`, to be moved by users needing non-default v4l2rtspserver options to `config/rtspserver.conf`

Tested and verified working.  Ready to merge.

tthk:
Note that this Pull Request supersedes Pull Request #170, and has fix for rtspserver.conf same as Pull Request #184.

shinebar1001:
Can you please update the documentation also?

tthk:
Sure.  Couple of questions about procedure:
Which documentation location is most appropriate?
Should I create another branch and pull request or add to documentation in this branch?

PR #212, opened 2018-04-08
Add picture timelapse capability

ping:
Adds timelapse script to save a snapshot after every N seconds (configurable via ``config/timelapse.conf`` or via Status page)


EliasKotlyar:
Many thanks @ping !

PR #216, opened 2018-04-09
Motion tracking

nik0:
Simple Motion tracking ...
Hope everything will work, sorry for my bad script and html ... Seems to work at home ....


EliasKotlyar:
Many thanks @nik0 !

Issue #30, opened 2017-12-28
Enable Full HD H264 stream 


EliasKotlyar:
According to the Datasheets from docs.7zip, the camera needs some special memory alignment to create the stream. 

```
3. Reserved memory and kernel reduction

Reserved memory is divided into two parts: ispmem and rmem:

     ispmem is the required memory for the ISP. The calculation method is that the Sensor outputs an image of length x width x4, for example, 1280x720x4 = 3600KB for a 720P Sensor and 1280x960x4 = 4800KB for a 960P Sensor. If you need to turn on the WDR feature, ispmem needs twice the size (length x width x4x2).
     rmem System Memory required for the Video Buffer, 720P probably need 14MB of memory, 960P probably need 19M memory.

The kernel default config to leave some margin, and the actual products often need to be cut to release more kernel space. The following are some of the commonly used options and explanations available for reduction:
```
This means, that for FULLHD the sizes should be:
ispmem: 1920*1080*4 = 8294400 = 8,2 MB
rmem = Should be somethere like 30MB

Unfortunately, the camera has the following settings:
ispmem=8M@0x6800000 rmem=16M@0x7000000
Source : https://github.com/EliasKotlyar/Xiaomi-Dafang-Hacks/blob/master/hacks/getroot.md

For now on, there is no way that the camera would stream h264 FULLHD....

EliasKotlyar:
Okey, i managed to start a FULLHD Stream using the following bootcmd:

```
setenv bootargs console=ttyS1,115200n8 mem=87M@0x0 ispmem=9M@0x5700000 rmem=32M@0x6000000 rootfstype=squashfs root=/dev/mtdblock2 rw mtdparts=jz_sfc:256k(boot),2048k(kernel),3392k(root),640k(driver),4736k(appfs),2048k(backupk),640k(backupd),2048k(backupa),256k(config),256k(para),-(flag)
```

Unfortunately, i dont know a way how to override this value without touching the bootloader

EliasKotlyar:
Okey, for the brave folks: (it can brick your camera - no unbrick possible!!!):

1. You need to dump the uboot-loader. Run the following to do this:
```
dd if=/dev/mtdblock0 of=uboot.bin
```
2. Open up the uboot.bin file with an hex editor, and change the "start-line" from 
```
mem=104M@0x0 ispmem=8M@0x6800000 rmem=16M@0x7000000
```
to
```
[...]

PR #218, opened 2018-04-10
Howto unbrick

lukeIam:
Some instructions how to unbrick the camera based on my experiences.
(I was able to unbrick my camera with those offsets)

### DIFF
--- a/hacks/unbrick.md
+++ b/hacks/unbrick.md
@@ -1,0 +1,46 @@
+## How to unbrick
+
+You need to open the camera and solder three pads (not too hard).
+And you will need a serial (usb) adapter to connect the camera to your pc.
+
+- Follow this steps to connect to the camera via serial connection:  
+https://github.com/EliasKotlyar/Xiaomi-Dafang-Hacks/blob/master/hacks/serial.md
+- If you are lucky you will see some status lines and a command prompt:
+`isvp_t20#`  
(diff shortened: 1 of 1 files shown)


EliasKotlyar:
Many thanks @lukeIam !

saipsa:
@lukeIam Will this work if I have bricked uboot? can I flash uboot.bin using this method?

lukeIam:
> Will this work if I have bricked uboot? can I flash uboot.bin using this method?

Hi @saipsa,

I haven't tried - but an long an your uboot is still booting you could give it a try.

saipsa:
@lukeIam Actually my uboot is not booting. 

I bricked while flashing uboot.bin https://github.com/EliasKotlyar/Xiaomi-Dafang-Hacks/issues/30

It said 513+0 records in but 512+0 records out, and something along the lines of,

memory full/no space available. cannot write.

So according to my understanding uboot is not booting. 

---

The status lines and command prompt will appear only if uboot is booting, and I can't use your method, if uboot is not booting? 

I am asking as I will have to purchase the FTDI adapter and soldering equipment.

EliasKotlyar:
@lukeIam : I tried this unbrick instructions today, and was able to recover my dead dafang. There was only a small change needed to make it work. Great work, thanks!

ronnyandre:
What to do if the serial doesn't read anything?

PR #221, opened 2018-04-11
Lighttpd

nik0:
- Uptodate webserver
- Choose the webserver to start in run.sh
- There are instructions to add login/password in files


EliasKotlyar:
Many thanks @nik0 !

jmtatsch:
Hi nik0,

I also started to use lighttpd now and noticed that it complains about not being able to write the access log.
`WARNING: unknown config-key: accesslog.filename`
According to https://unix.stackexchange.com/questions/11788/why-is-lighttpd-saying-unknown-config-key-accesslog-filename we would need to add "mod_accesslog" to the modules.
I tried that, however when doing that lighttpd doesn't start because it can't find mod_accesslog
```
2018-05-05 05:26:03: (plugin.c.170) mod_accesslog  plugin not found 
2018-05-05 05:26:03: (server.c.1142) loading plugins finally failed
```

Could you compile lighttpd again with that module enabled? Or did you disable that on purpose?

nik0:
See https://github.com/EliasKotlyar/Xiaomi-Dafang-Hacks/pull/328
I also add ssl module, so you need to have some new libs from libs (and now there is lighttpd that start lighthttp.bin)

jmtatsch:
Great, we can really consider dropping boa support now. I mean it hasn't be developed since 2005.

PR #226, opened 2018-04-13
Jpeg optim

nik0:
- For emails as the pictures were huge
- Experimental for cgi currentpic.cgi, try "currentpicoptim.cgi" that use some CPU and may have the camera not stable (to try)


shinebar1001:
Thanks!

PR #232, opened 2018-04-15
lighttpd.conf: Make default port 80, add mimetypes for js and css

ping:
Since lighttpd is an alternative to boa, I thought it made more sense that it should also be listening on port 80 instead.

The config was also missing the mimetypes for css and js.

An additional aside,  I noticed in ``/var/log/error.log`` there appears to be a warning with the default lighttpd config:
``
(server.c.1452) WARNING: unknown config-key: accesslog.filename (ignored)
``
I did not fix it in this PR since it's just a warning and the required module may be added in a future update.

### DIFF
--- a/firmware_mod/config/lighttpd.conf
+++ b/firmware_mod/config/lighttpd.conf
@@ -1,12 +1,14 @@
 server.document-root = "/system/sdcard/www" 
 
-server.port = 7000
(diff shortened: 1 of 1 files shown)


EliasKotlyar:
Many thanks @ping ! I have merged this change.

PR #233, opened 2018-04-15
minor changes for more consistency in mqtt topics, add home assistant integration tutorial

jmtatsch:
had to add the home assistant integration tutorial here because unfortunately non code owners apparently cannot collaborate on wiki stuff.
We should think about merging the wiki contents back into the markdown collection here so there is just a single point of truth.


EliasKotlyar:
Hello jmtatsch,

Many thanks for your contribution. I have merged it to the master branch.

Greatings
Elias

Issue #217, opened 2018-04-09
New UI


gysmo38:
Hello,
I start to write a new html5 ui. I give you, if soemone want to try and tell me what to improve.

It is not finish, I hope to finish it in few days.

Just create a directory ui under www directory and unzip the file.

[ui.zip](https://github.com/EliasKotlyar/Xiaomi-Dafang-Hacks/files/1891753/ui.zip)

fcastroruiz:
its good option , work fine! some links dont work fine (normal in beta version) but i think is great job

Psudonico:
good work,
I'am colorblind and do not distinguish the active buttons from the non-active ones

jmtatsch:
This color schema is very hard to read
<img width="417" alt="screen shot 2018-04-13 at 08 00 52" src="https://user-images.githubusercontent.com/3080674/38720953-8643a2fe-3ef8-11e8-870a-c037ad2ccecf.png">

ping:
My apologies if this is bad form, but I am also working on a new UI.
Latest changes are on my fork at https://github.com/ping/Xiaomi-Dafang-Hacks/tree/new-ui where you can download and replace your copy of the ``www/`` folder.

The new UI is responsive and should work mostly fine on both mobile and desktops. Comments/feedback are welcomed.

### Screenshots

![newui-1](https://user-images.githubusercontent.com/104607/38770054-71d78978-403f-11e8-88e2-fd26efbb8d3b.png)

![newui-2](https://user-images.githubusercontent.com/104607/38770056-8019f82c-403f-11e8-9547-08848b270a91.png)

![newui-3](https://user-images.githubusercontent.com/104607/38770058-891e3a32-403f-11e8-87d7-5c919a59e9ca.png)

remb0:
Nice,  i hope or wilill be the default.

desq42:
+1

jmtatsch:
@ping Love your work, would you maybe open up a pull request, so we can continue discussing things there?

ping:
@jmtatsch PR submitted

jmtatsch:
pr #239 was recently merged, please continue bringing in your great ideas

PR #239, opened 2018-04-17
[WIP] New UI

ping:
This is the PR for the UI work I mentioned in https://github.com/EliasKotlyar/Xiaomi-Dafang-Hacks/issues/217#issuecomment-381340148

Change summary:

- Incorporate the [Bulma CSS framework](https://github.com/jgthms/bulma)
- Changed many of the functions in ``action.cgi`` ``scripts.cgi`` ``status.cgi`` ``getMotionInfo.cgi`` to return json so that we can have more elegant ajax-y UI flows
- Changed the "live" view to only refresh statuses on demand (when Camera Controls menu is hovered/clicked) instead of a constant interval refresh. I did not like the constant refresh in the current implementation because it was sometimes locking up the UI in the background.
- Changed the "live" view image to reload only after previous image is loaded instead of a constant refresh interval (uses image "onload" to trigger)
- I've fixed up some bugs in configMotion.html and also made the UI work responsively
- Added a "shutdown" command (calls ``/sbin/halt``) so that I can gracefully shutdown the camera instead of pulling the plug
- Add option to choose a dark UI theme

This is tentatively still a WIP because I feel that more testing is required but I've submitted this PR as a basis for discussion/feedback since I've had to change many of the core .cgi scripts.

Edit: Latest screenshots
gh_dafang_01
gh_dafang_02
gh_dafang_03
gh_dafang_04
gh_dafang_05
gh_dafang_06


firmware_mod/www/cgi-bin/action.cgi
```diff
@@ -121,45 +127,45 @@ if [ -n "$F_cmd" ]; then
...
       if [ "$(cat /etc/TZ)" != "$tz" ]; then
-        echo "Setting TZ to '$tz'...<br/>"
+        echo "<p>Setting TZ to '$tz'...</p>"
         echo "$tz" > /etc/TZ
-        echo "Syncing time...<br/>"
-        if [ "$(/system/sdcard/bin/busybox ntpd -q -n -p $conf_ntp_srv 2>&1)" -eq 0 ]; then
-          echo "<br/>Success<br/>"
-        else echo "<br/>Failed<br/>"
+        echo "<p>Syncing time...</p>"
+        if [ "$(/system/sdcard/bin/busybox ntpd -q -n -p time.google.com 2>&1)" -eq 0 ]; then
```
jmtatsch: this may break ntp configuration via file?

ping: @jmtatsch Fixed. Thanks for the catch. Turns out that I missed out on the ntp form in the ``status.cgi`` as well.

gysmo38:
I test your UI, it is so great :) better than mine :p 
Some comments:
- I find better to have ptz button (left, right, top , down) directly on picture
- in camera controls the button not reflect the status of the camera
- on my laptop the picture is little to be so I need to scroll down to have view it
- maybe add camera name in title could be great

If I can help let me knwon

jmtatsch:
I would also love gysmo’s on picture motion controls in the mix

gysmo38:
@ping 
- great :)  maybe if it possible to put the pad at right bottom corner. Like in tiny cam :p 
- I made more test with button in camera controls. All button are off for me. If I toggle one, the command work but after 1 or 2 seconds it come back to off. I check, I have no javascript error.
- Thank the live view is better. I have scrolldown for the bottom page. Is it possible to reduce space at the top to have full page on the screen?
- great :)

ping:
@gysmo38 
- Never used tinycam so I've no idea about their UI. It's difficult to pin the dpad flushed to the side of the image because the image is responsive. If you know how to, please let me know.
- Can you check a few things? Open up the Developer Console => Network tab, then do a hard refresh (control+F5 I think, or control+click on refresh). Do you see the requests being made below?
Does opening http://<camip:port>/cgi-bin/state.cgi?cmd=yellow_led in a new tab give the correct "on"/"off" response?
I've patched the cgi scripts to always return no-cache in case browser caching was causing problems, so please refresh your version again.

dafang_debug_network

MkLHX:
Hi there!
I'm propose an other ui for only html files in the project by using https://getmdl.io/ framework

So i submit a PR #254 

If i can help for a new responsive ui i'm in!

EliasKotlyar:
Hello everyone,

Many thanks contributions for the new UI. I have tested it, and its looking great. :) 

Greatings
Elias

gysmo38:
@ping I tested with the commit. It work for me :)
I will try to move pad. 
Another idea, for rebooting have a coutdown (maybe 30s) before refresh page.

PR #241, opened 2018-04-17
Fix Stuck in Alarm

Paxy:
Fix Stuck in Alarm  state if $save_snapshot is false


EliasKotlyar:
Hello  @Paxy ,

many thanks for your contribution!

Greatings
Elias

PR #244, opened 2018-04-18
Autoupdate script

nik0:
This script should auto update from github.
Try with `autoupdate.sh -d /system/sdcard -x -p -v` to see what will happen, and then remove -p to really execute the update.

**Be carrefull of your config file they can be erased !!!**

### DIFF
--- a/firmware_mod/autoupdate.sh
+++ b/firmware_mod/autoupdate.sh
@@ -237,7 +237,7 @@ if [ $_FORCE = 1 ]; then
 fi
 
 if [ $_PRINTONLY = 1 ]; then
-    log "Print action, do nothing"
+    log "Print actions only, do nothing"
 fi
 
 if [ ${update} = true ]; then
--- a/firmware_mod/autoupdate.sh
+++ b/firmware_mod/autoupdate.sh
@@ -134,16 +134,21 @@ isNeedtobeUpdate()
 # Recursive call (for folder)
 getfiles()
 {
-    for row in $(echo "${1}" | ${JQ} '.[]| select(.type=="file") | .download_url' ); do
-        filetoget=$(echo "${row}" | tr -d '"')
-        echo ${filetoget}
-    done
+    if echo "${1}" | grep -q "API rate limit exceeded"; then
+        logerror "Github limit exceeded, try with an account (-u option)"
+        exit 1
+    else
+        for row in $(echo "${1}" | ${JQ} '.[]| select(.type=="file") | .download_url' ); do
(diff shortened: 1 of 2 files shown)


msj33:
Yes - Very cool idea.

Could this be presented in the WEBUI with 3 options:

-Disabled
-Manual - WEBUI gets a update notification, but update button should be used.
-Automatic - Updates are installed automatically(Incl. reboot).

I think automatic should be the default config - Reason: 
- General community is on latest version(Removes reports of already fixed bugs)
- Faster and more feedback on latest features/bugs.

The only issue/risk I see around this is lack of brachning/versioning on the repo.

nik0:
Be carrefull with the automatic mode, I am not sure everything is configured throught conf file.
For example I have customized boa port and login/pass and copying the boa.conf from repo is not a good idea ..

msj33:
Agree - The repo files needs to be isolated from user-specific config - WPA conf is same issue.

jmtatsch:
Wouldn't it be cleaner to

1. to compile git for the platform
2. to move the contents of firmware_mod to a new "Xiaomi Dafang OS" repo
3. then just trigger git pulls via ui/automation?

nik0:
Using git was my first thought:
- It have been compiled in the source repo
- But you are right the repo would have need to be changed and the .git folder takes some place ....

jmtatsch:
If we have a working git binary, the size of the . git should not be an issue.
We could perform a shallow checkout with `git clone --depth n` with a history truncated to the specified number of revisions. Also there is `git gc` and `git prune`.

Maybe we can get @EliasKotlyar opinion about splitting the repo.

I would propose the following split:

Xiaomi-Dafang-Hacks (the hacking information, building scripts and custom firmware )
Xiaomi-Dafang-Software (the former firmware_mod folder)
Xiaomi-Dafang-Software-Development (the toolchain setup & all the build recipes)

EliasKotlyar:
Hello everyone,

@nik0 : Many thanks for your contribution. I have tested it, and its looking good.

@jmtatsch : I have started working on a git binary. Unfortunately i cant get it compiled yet, because of some dependency issues(libcrypto). I would not prefer the split into 3 different repos because its really hard already to manage 2 of them properly. I think it should be easy to achieve a update of only one directory(sparse-checkout). Refer to this stackoverflow post:
https://stackoverflow.com/questions/600079/how-do-i-clone-a-subdirectory-only-of-a-git-repository/13738951
I havent tried it yet, but in my opinion this should be possible. I will try it out as soon as i get the git binary running.

Greating
Elias

Issue #259, opened 2018-04-20
New GUI motion configuration


nik0:
The motion configuration page is not working anymore:
- motion detection region select is not working at startup (need to enable/disable) to activate the region
- invalid argument when press the "configuration motion" button

Idea of improvement for the logs: have a summary is good but maybe we can propose to download the files ...

Anyway nice work !!!

Nik0

nik0:
This still not working for me :-( The first load of the page doesn't display the region.

Apparently I never get into  the done function below (I added a log that nevers happens in console)
```
 $.ajax({ 
            'url': '/cgi-bin/getMotionInfo.cgi',                          
            'dataType': 'json'                                                     
        })                                                                               
.done(function (motionInfo) {  
```
One more thing: I guess the resize of the image is not well managed with imgAreaSelect. According to the documentation ( http://odyniec.net/projects/imgareaselect/usage.html ) the properties imageHeight and imageWidth have to be set if the image is scaled.

Thks !

ping:
It's my bad, I didn't include `getMotionInfo.cgi` in the new UI PR, which probably explains your issues.

Can you grab the changed files (`www/cgi-bin/getMotionInfo.cgi`, `www/configmotion.html`) from
https://github.com/ping/Xiaomi-Dafang-Hacks/commit/8ec2218d13611479f72012af3ebe4b08b9af2ed2 and replace your copy for testing?

nik0:
Good it is working now !!!
I will never find the issue by myself ...
Thanks !

PR #260, opened 2018-04-21
Fix motion configuration UI

ping:
Re #259 

Changes:
- load the css/js files needed for config motion in index.html so that we don't have to load these assets dynamically from the page. Since these are small and served locally, I don't see any major impact from this change.
- I could not replicate the "invalid argument" error as reported in the issue but there was a bug with saving the "Enable motion tracking" which is fixed in this pr

### DIFF
--- a/firmware_mod/www/configmotion.html
+++ b/firmware_mod/www/configmotion.html
@@ -74,65 +74,54 @@
 
     }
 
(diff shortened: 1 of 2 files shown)


jmtatsch:
thank you

PR #261, opened 2018-04-21
Fix rtsp-h264 state response

ping:
This was causing the status of "RTSP H264" to be wrongly reported in the UI.

We cannot use ``rtsp_server`` from ``common_functions.sh`` because that is non-specific to h264, if either rtsp-h264 or rtsp-mjpeg is activated, ``rtsp_server`` reports as "ON"

Regression introduced in #207


jmtatsch:
Hi Ping,

I took the liberty to tackle this one level deeper in common_functions.
Thanks for pointing this out and offering a quick fix. State response works for me now, please let me know if it still doesn't work for you.

PR #263, opened 2018-04-21
reworked timelapse script into a simple job control script

th0ma5w:
This pull request modifies the "timelapse" web based toggle control script to start and stop the existing script "timelapse.sh" according to its conventions. It provides the following fixes:

- Start and stop now works as intended by the web interface
- Uses existing configuration parameters
- Can stop an automatically started script (not tested, but should work?)
- Displays the process ID correctly in the web interface

### DIFF
--- a/firmware_mod/controlscripts/timelapse
+++ b/firmware_mod/controlscripts/timelapse
@@ -1,38 +1,28 @@
 #!/bin/sh
+PIDFILE='/run/timelapse.pid'
 
 status()
-{  run_timelapse=$(cat /run/timelapse.status)
-   if [ "$run_timelapse" == true ]; then
-   echo "started"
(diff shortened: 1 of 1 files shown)


ping:
Wow, I didn't realise that the original timelapse controlscript from #212 never got added to the repo even though the PR got merged. That's super bizarre.

It will be good if the start function checks if an existing process is already started before starting timelapse.sh

```
start()
{
	if [ -f "$PIDFILE" ]; then
		echo "Timelapse already running";
	else
		echo "Starting Timelapse"
		/system/sdcard/bin/busybox nohup /system/sdcard/scripts/timelapse.sh &>/dev/null &
		echo "$!" > "$PIDFILE"
	fi
}
```

PR #267, opened 2018-04-21
More fixes for config motion (fixes #259 finally)

### LINKED ISSUE #259: New GUI motion configuration

ping:
Changes:
- Add configMotionInfo.cgi that was missed out in the new UI PR
- Do manual scaling for imageareaselect


jmtatsch:
Thank you

PR #269, opened 2018-04-21
Add more log display on interface

nik0:


jmtatsch:
Great idea. Thank you.

PR #270, opened 2018-04-21
Video server monitoring

nik0:
Monitoring and log for RTSP process 
I suspect this process to crash, so monitors it and get the logs
If this method is OK for everyone we can add it to rtsp-mjpeg


jmtatsch:
I really like the idea of getting the log file. But is it really crashing so often we need another running loop watching something? Or is the log growing so rapidly?

nik0:
I am not sure about the crashing. It seems that sometime I don't have any stream anymore but I didn't investigate more. This is more a tesing PR, we don't need to add this loop for now.
For the logs I didn't let my camera running long enough to set it ...

PR #277, opened 2018-04-22
rename wpa_supplicant.dist to prevent mistakes

mookie-:
Hello,

I was stupid enough to do a `mv wpa_supplicant.dist wpa_supplicant` (just removed the .dist) and was two days busy to figure out how to get the camera working :)

I think it would be better to name the file `wpa_supplicant.conf.dist` but if not, just close the PR.
Thanks to all for the nice work you did!


jmtatsch:
.dist files are a unix convention see e.g. https://stackoverflow.com/questions/16843080/what-does-dist-used-as-an-extension-of-some-source-code-file-mean/16843246

Dopeyr:
To be fair, the stack overflow link does say:

>So, to get started with such frameworks, **you should remove the .dist extension**, and customize your configuration file with your personal parameters.

So I think this PR should at least be considered, as removing .dist from wpa_supplicant.dist results in a non-working config, whereas removing it from wpa_supplicant.conf.dist does not. Well, as long as fill in the right values :)

However, it should probably apply to other .dist files too.

jmtatsch:
I understand and agree now. We will have to change it for everything though to make it consistent.

mookie-:
I've made a new commit with all dist files. https://github.com/mookie-/Xiaomi-Dafang-Hacks/commit/573422eaa5cd15d4deb25704430208dc62d896bb
It seems github doesn't update this PR. Should I create a new one?

jmtatsch:
Probably because this one was closed prematurely. You can reopen or start a new one. Think also of the gitignore files and the comments within the dist files telling what to do.

jmtatsch:
And also the install_cdw.md

PR #278, opened 2018-04-23
adding tips from Issues replies

JayCuthrell:
Adding a few tips that I found to be successful from the Issues replies


jmtatsch:
Thank you for your contribution

PR #279, opened 2018-04-23
rename .dist files to .conf.dist

mookie-:
Continuing on #277 .

> Think also of the gitignore files and the comments within the dist files telling what to do.
> And also the install_cdw.md

The gitignore and comments don't have to be changed because the original filename will still be the same.

I've updated all *.md files where I could find a `.dist`.


jmtatsch:
That is true. Thanks for your contribution.

PR #281, opened 2018-04-23
Use retain flag for MQTT topics

### LINKED ISSUE #280: Automatic Discovery Publishing
It would be super helpful if discovery happened routinely. 

The discovery support for Home Assistant is SUPER helpful, but it is lost every time Home Assistant is restarted, which ends up being frequent for those that work on their system a lot.

It could even be nice if there was an endpoint to hit to trigger discovery, then maybe we could hit it when Home Assistant finishes booting.

maddox:
This adds the use of the `retain` flag on the published MQTT topics. 

The retain flag tells the MQTT server to keep the state local in case a client makes a request on those topics. This means clients can fetch state without waiting for a publishing event. This is especially helpful for things like Home Assistant, which will look up the topics to fetch state after starting.

Right now, if you restart Home Assistant, you lose all of the auto discovered components that this project provides it. To get them to show up again in Home Assistant, you have to restart the camera or manually restart the MQTT control service and status service. 

With this patch, none of that is needed anymore. Since the state will stay on the MQTT server, Home Assistant will be able to have it on boot and automatically add all of the components and their last state. 

This resolves #280


jmtatsch:
Love it. Thank you very much.

Issue #79, opened 2018-02-09
Light sensor problem


hondaj:
when the light in the room is on the switching threshold, the camera constantly switches between the day / night modes. Is it possible to set the "hysteresis" somewhere in the configuration files? Is switching support embedded in the firmware?

This is a serious problem when you have motion detection enabled by the NAS software. And in addition, it can be a potential cause of damage to the filter switching system, which is implemented mechanically.

shinebar1001:
A Hysteresis is not implemented now. One could create an average for 2 to x cycles to do this. That would fit to situations were the light smoothly dims, but not to situations where you Switch of a light in a cellar, so this would be another Option for the light sensor.

The switching support is quite unsophisticated: it takes 20 Bytes from the sensor and Counts the dots, as I didn't manage to fiddle out how the sensor data is formatted. So you could try to set the threshold higher in line 11 of scripts/ldr.sh just Change the 50 to 55 up to 60. But highly possible that won't fix your Problem.

shinebar1001:
See #85 for a fix - this implements a configurable averaging for the light-measurement, so the constant switching should disappear, when you set the value > 1

hondaj:
Thanks for add averaging and simple managing over web. I played with sleep time in previous of ldr.sh setted to 2m, 5m. I has a good result when light smoothly dimms, but has a one drawback. At night, when you turn off the light in the room, you have to wait a few minutes to switch to night mode :-( I think it should be added second mechanism whitch check "extreme" light change (if it's overall possible) and switch immediately when "detects" peak of light or dark. Today I updated my files from your latest work (thank you once again), seting AVG=3 and test it now for a 1-2 days. I give you an answer how it works in my conditions.

shinebar1001:
The easiest way would be to reduce the delay between the measurements. (last but one line in /scripts/ldr.sh) - change sleep 30 to let's say sleep 10

Considering the detection of extreme light changes: I've got to think about that, but it's not really easy. Probably you got an Idea.

Let's say we go with 5 data points:
0,0,0,0,0 now the light is switched of, so we get 0,0,0,0,60 
my first idea was to compare the old average to the newest measurement:

if average of old measurement < 21 and new measurement = 60 => disable IR
This would mean:
0,0,0,0,0 new 60 -> 0 < 21 & new = 60 -> disable IR
0,0,0,0,60 new 60 -> 12 < 21 & new = 60 -> disable IR
[...]

PR #305, opened 2018-04-26
Added nightmode.sh script (to be called by the new autonight binary) …

jthwho:
This adds the scripts for issue #79.  I've also submitted a pull request to the Software repository containing the source for the binary application that'll need to go along with this.

### DIFF
--- a/firmware_mod/controlscripts/auto-night-detection
+++ b/firmware_mod/controlscripts/auto-night-detection
@@ -1,5 +1,6 @@
 #!/bin/sh
 PIDFILE="/run/auto-night-detection.pid"
+CONF_FILE="/system/sdcard/config/autonight.conf"
 
 status()
 {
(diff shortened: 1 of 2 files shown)


jthwho:
Yea, the binary has what I think are reasonable defaults for now.  I'm still tweaking the parameters (I think I can get it to respond much better to quick lighting changes).  Once I've finalized on some parameters I'll submit a pull request with an autonight.conf.dist containing those values (which I'll also update in the binary).

Also, I should note this pull request doesn't have the autonight binary (since I expect it'd probably be bad to have random strangers checking in binaries).  I'm under the assumption that the autonight binary will get built by someone trusted and included when this pull is merged.  Is that right?

jmtatsch:
Couldn't agree more on the binaries from strangers thing, we really have to be careful there.
Thank you very much for your contribution.

jthwho:
No problem.  Happy to contribute.  The flexibility of the camera combined with this project is great.  Thanks for sharing it with us.

Issue #291, opened 2018-04-24
Updating Firmware Killed Dafang


mynameisdaniel32:
I just had a go at updating my firmware (was previously using one from about a month ago) and now the camera won't connect to WiFi at all. When powering up I get a solid orange LED for about 5 seconds, a half second blink of the blue LED, then it goes back to the orange LED.

I have tried reflashing the CFW without change. I formatted  the entire SD card and put a fresh copy of the firmware files onto it (and recreated the wpa_supplicant.conf file) without improvement.

Any ideas?

jmtatsch:
Maybe your motors are stuck. Try commenting out hcalibrate & vcalibrate in run.sh

jmtatsch:
And please rename you issue, it is highly misleading.

jmtatsch:
Probably related to #287

mynameisdaniel32:
Changed the title. I don't think it's related as my camera won't even connect to WiFi for me to ping/telnet to it.

I also spent some time reverting the firmware to what was on there previously, as well as an older build from here, both now have the same issue so something isn't happy. Will try stripping down run.sh to see what happens (motors were working fine when I reflashed the cfw though).

mynameisdaniel32:
I've worked this out and resolved the problem. Turns out when I updated the firmware using the GitHub Desktop software to clone the repo, it does something strange to the files and they cause corruption when read by the camera. Downloading the zip and using the files from that was my only change to fix it.

I only just found this from a [comment](https://github.com/EliasKotlyar/Xiaomi-Dafang-Hacks/issues/197#issuecomment-379166400) somewhere deep in another issue, so I'm leaving this open for now as it's probably worthwhile adding a warning to the installation instructions? Or even better investigating the cause of this and fixing it.

Dopeyr:
Perhaps a `.gitattributes` file in the repo would resolve this? I haven't looked into it in any depth but https://help.github.com/articles/dealing-with-line-endings/#per-repository-settings suggests `.gitattributes` could control the line endings?

atus:
@mynameisdaniel32 it's the line endings in windows vs linux. The camera can't digest \r\n. That's why the install instructions say download instead of clone repo.

sajeevfive:
I have the same problem but I did firmware update through Mobile app.

Can someone provide the steps to solve this issue as I'm a newbie.

Thanks in advance, saj

PR #309, opened 2018-04-28
Always checkout .sh and .dist with UNIX line endings

Dopeyr:
Ref #291 Always checkout `.sh` and `.conf.dist` with unix line endings

This should help resolve issues when cloning on windows.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,3 @@
+* text=auto
+*.sh text eol=lf
+*.conf.dist text eol=lf


Dopeyr:
Hopefully that's got them. I ran:

`find firmware_mod -type f -exec file {} \;|grep 'executable.*ASCII'`

to find them. Most are covered by .sh, .cgi. 

I just need to check on windows that this actually resolves the issue.

jmtatsch:
That would be great. Let me know how it goes then we will merge it.

Dopeyr:
Looks ok here. I've cloned the repo from scratch on windows and the expected files have unix line endings.

PR #319, opened 2018-04-30
use relative paths for web interface.

gkaindl:
Updated the web interface to use relative paths, rather than (hardcoded) absolute paths.

This is useful if you want to run your camera(s) behind a reverse-proxy, setting it up like "myserver.com/cam1", "myserver.com/cam2", and so on.

Since the html isn't generated dynamically, I think that's the most straightforward way to achieve this, without breaking the normal usecase.


peturdainn:
nice! I was about to do the same thing, because I indeed intend to run it behind a reverse proxy.

jmtatsch:
Thank you for your contributing.

Issue #249, opened 2018-04-19
Increased power consumption after using up, down and left movement


peturdainn:
As part of testing this for installation, I did some power consumption tests, and noticed something strange:
(movements done using live view web page arrow icons, consumption is *after* motion finished of course)

After boot, 3.8W
Press LEFT: 5.3W
Press RIGHT: back to 3.8W
Press UP: 6.8W
Press LEFT: 8.3W
Press DOWN: 6.8W
Press RIGHT: 5.3W

so different direction cause different extra consumption, and after using up or down the initial 3.8W is never possible any more (where the extra from LEFT is gone again after RIGHT)

I have not yet checked the code myself so I may update this issue if I find out more.

peturdainn:
Had a quick look in /drivers/misc/sample_motor/motor.c and after moving the gpios aren't switched off? 

Looks like it's doing microstep (8 steps), the driver should store its sequence in the steps but turn off all windings when target is reached. You don't want to power the windings to hold position.

But this doesn't explain why there is a difference between left/right/up/down :(

hberntsen:
Same issue here, running `/system/sdcard/bin/motor -d s` reduces the power usage to a normal level again which is between 1.8-2.2W in my case.

peturdainn:
@hberntsen if you want to test my fix for the motor power usage, I've attached the driver.
With this my idle current is always below 2W

[sample_motor.zip](https://github.com/EliasKotlyar/Xiaomi-Dafang-Hacks/files/1959276/sample_motor.zip)

peturdainn:
There seems to be an identical issue with the IR cut filter. I think the gpio only needs activating to switch.
Sadly the state of the gpio is used for status

peturdainn:
@EliasKotlyar the IR Cut status is using the gpio to report on/off, and I want to disable both gpio after switching.

What would be an acceptable way to store the state for you? A file on /tmp?

hberntsen:
@peturdainn Just tested your driver, can confirm it works! Power usage drops back to normal indeed. I also have the issue with the IR cut filter. After booting I see a power usage of 1.15W, after changing the IR cut status it is 1.40W.

peturdainn:
great, thanks for confirming!

I'm now beginning to fear that the reports about broken camera's (too much power draw, etc) are related to the gpio of motor and ircut staying on. This extra current draw is not healthy.

peturdainn:
[...]

PR #325, opened 2018-05-01
only toggle gpio for ircut, keep state in tmp file

peturdainn:
As mentioned while investigating/discussing issue #249 there is also a problem with the gpio of ircut staying active.

This PR changes this into a toggle, the gpio is back to zero after toggling.

An additional problem is that the state was using the gpio state, this has been changed by using a temp file containing 0 or 1, it is written in /var/run/ircut  (better locations welcome)


EliasKotlyar:
Hello @peturdainn ,

Thanks your for your contribution! I have merged it.

PR #335, opened 2018-05-07
Before setting http password url decode it

tssva:
Decode the url encoded http password before setting it, so passwords with special characters are properly set.


jmtatsch:
Thank you for your contribution

PR #338, opened 2018-05-09
Remove dropbear key. A unique one will be created at runtime.

tssva:
Dropbear is launched with the -R option, so if a host key doesn't exist dropbear will create one when the first ssh connection occurs. Having the key in the repo prevents this from occurring causing all devices to have the same ssh host key which is poor security practice.


jmtatsch:
Good idea. Thank you.

Issue #337, opened 2018-05-08
v4l2rtspserver problem


khseal:
It runs without problems, but there is no picture on the ip address.
When you click the start button in the services. The button is spinning endlessly.
When starting from the console, everything works
/system/sdcard/bin/v4l2rtspserver-master -S -W1600 -H900

khseal:
Logs after stratup and try connect vlc
LAME 3.100 32bits (http://lame.sf.net)
Using polyphase lowpass filter, transition band: 16538 Hz - 17071 Hz

[NOTICE] /home/osboxes/SRC/Xiaomi-Dafang-Software/v4l2rtspserver-master/src/main.cpp:646
	Create ALSA Source.../dev/dsp
[NOTICE] /home/osboxes/SRC/Xiaomi-Dafang-Software/v4l2rtspserver-master/src/ALSACapture.cpp:51
	Open ALSA device: "/dev/dsp"Changed OSD

[NOTICE] /home/osboxes/SRC/Xiaomi-Dafang-Software/v4l2rtspserver-master/src/DeviceSource.cpp:87
	begin thread
[NOTICE] /home/osboxes/SRC/Xiaomi-Dafang-Software/v4l2rtspserver-master/src/ALSACapture.cpp:68
	Channel Count:1
[NOTICE] /home/osboxes/SRC/Xiaomi-Dafang-Software/v4l2rtspserver-master/src/DeviceSource.cpp:87
	begin thread
[NOTICE] /home/osboxes/SRC/Xiaomi-Dafang-Software/v4l2rtspserver-master/src/main.cpp:151
	Play this stream using the URL "rtsp://192.168.5.39:8554/unicast"

[NOTICE] /home/osboxes/SRC/Xiaomi-Dafang-Software/v4l2rtspserver-master/src/ServerMediaSubsession.cpp:24
	Creating Source...

[NOTICE] /home/osboxes/SRC/Xiaomi-Dafang-Software/v4l2rtspserver-master/src/ServerMediaSubsession.cpp:50
	Creating Sink for format : video/MP2T
[NOTICE] /home/osboxes/SRC/Xiaomi-Dafang-Software/v4l2rtspserver-master/src/ServerMediaSubsession.cpp:24
	Creating Source...

[NOTICE] /home/osboxes/SRC/Xiaomi-Dafang-Software/v4l2rtspserver-master/src/ServerMediaSubsession.cpp:41
	Creating normal Source...
[...]

PR #339, opened 2018-05-09
Remove quotes around options or v4l2rtspserver-master does not start

Dopeyr:
Ref #337, remove quotes or v4l2rtspserver-master does not start on boot or via the UI.


jmtatsch:
Thank you for fixing this.

PR #340, opened 2018-05-10
Create root home directory and etc directory on sdcard

tssva:
1. If they don't exist `/system/sdcard/root` and `/system/sdcard/etc` directories are created.
2. Files needed from the original `/etc` are copied over to `/system/sdcard/etc`
3. `/system/sdcard/etc/passwd` is modified to make the home directory of the root user `/root`
4. `/system/sdcard/root` is bind mounted to `/root`
5. `/system/sdcard/etc` is bind mounted to `/etc`

End result is writeable /etc and /root directories, so things such as root user password changes and ssh key based login can now work.

### DIFF
--- a/firmware_mod/run.sh
+++ b/firmware_mod/run.sh
@@ -17,6 +17,25 @@ killall telnetd
 . /system/sdcard/scripts/common_functions.sh
 echo "loaded common functions" >> $LOGPATH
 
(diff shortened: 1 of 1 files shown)


jmtatsch:
Love it.
A big step towards better security :)
Thank you for contributing.

PR #345, opened 2018-05-10
Remove quotes around options for v4l2rtspserver

Kaeltis:
(same as d59e31d775170220b5af7c0938425adec2c5132b but for rtsp-mjpeg)


jmtatsch:
Thank you for catching that.

Kaeltis:
You are very welcome, thanks for all your (and the other collaborators) work on this :)

PR #354, opened 2018-05-13
Two ways audio (not to be merged)

nik0:
To be tested: new page to be loaded: audiooutput.html (not integrated for now in index.html)
Not sure out volume is working
I didn't tested with audio recording ...
**You need to start manually audioserver binary**

# Instructions for certificates
This feature use embedded audio features of moderns browsers ,but Chrome, and Firefox (not so restrictive) need https connections ! I tested this on latest version of Chrome and Firefox on Windows
You need to use self-signed certificate (unless you have a "true" certificate with a real domain name).
Follow the instructions there: https://community.ubnt.com/t5/EdgeRouter/WANTED-Instructions-for-self-signed-certificate-that-Google/m-p/1984552#M167185
(I copied the script in bin/createCertificate.sh) to be used as this:
`createCertificate.sh dafang.com` (dafang.com have no importance but need to be declared in /etc/hosts whatever the OS), then you will access to your camera with the url `https://dafang.com/audiooutput.html `

When the certificates are created (**it will override the existing lighttpd one !**) you need to insert them in your browser:
# Chrome:
Copy` /system/sdcard/config/lighttpd.pem` and `/system/sdcard/config/ubntCA.pem ` on your computer and insert them. **Make sure you select “Trusted Root Certification Authorities” as destination for your certificate**
https://www.nullalo.com/en/chrome-how-to-install-self-signed-ssl-certificates/2/

# Firefox:
- Preferences / Privacy and Security 
- Certificates / View Certificates and import the two certificates (select "Trust this CA to identify websites") (not sure the 2 certificates need to be imported)

Note that you can use this without all this certificate stuff on Firefox (use simple http connection).

# How does it work
It uses https://webrtc.org/ to get the audio, then use websockets to send the raw PCM to the camera. 
It connects to port 7681 for https (wss in websocket) and 7682 for http (ws in websocket). Theses ports can be changed in "audioserver" parameters.


iashwinprasada:
cool work, ill check this out.

nik0:
Will create a new one soon

kylerw:
was this merged or restarted elsewhere?

nik0:
No, not yet ... lack of time and with the new boot loader that won’t use the embedded audio driver some part need to be re-written

matchett808-gh:
I might like to pick this up - was this branch in a working state (ie - would a device call _actually_ work)?

(I'm not even sure _what_ the ecosystem here is - so I'm thinking a bit of rejigging will be easier for me to do lol)

PR #361, opened 2018-05-15
Remove "device_class": "sensor" from light sensor config

bieniu:
In Home Assistant 0.69+ there are four device_class for sensors: battery, humidity, temperature, illuminance. device_class sensor isn't allowed and cause error in mqtt component. This sensor value in %, which is why in my opinion it shouldn't have device_class (illumination class should has lm or lx unit).
https://www.home-assistant.io/components/sensor/#device-class


jmtatsch:
Thank you for catching that.

PR #368, opened 2018-05-18
Add compression and day folder to timelapse

mdlima:
Add 2 new options to timelapse.conf:

- SAVE_DIR_PER_DAY
if set to 1, will create a new subfolder for each day to save the images

- COMPRESSION_QUALITY
used with jpegoptim to add compression to the saved images

If no changes are made to the timelapse.conf file everything will work as before.


jmtatsch:

PR #371, opened 2018-05-22
Auto update2

nik0:
Update autoupdate script:

- remove some options, now compare always file by file
-  two stages update: copy the needed files in /tmp folder, stop all services bftpd and lightthttpd and reboot
-  No more restriction on non updated files, only *.conf and *.user are ignored


jmtatsch:
Very nice 👍 
I guess you checked in autoupdate.sh.orig by mistake?

nik0:
yes correct it was an old version of the script :-)

PR #373, opened 2018-05-23
Add an API call to trigger sendPictureMail

Dopeyr:
Added an api endpoint for triggering sendPictureMail. This is useful for me so I can get openhab to trigger when motion is detected via external sensors (not the motion detection on the dafang).

@kszere I'm not sure if this PR should be here, or in https://github.com/kszere/Custom-Software-For-Xiaomi-Dafang, or how https://github.com/kszere/Custom-Software-For-Xiaomi-Dafang gets back into this repo.

Any help appreciated :)


kszere:
@Dopeyr I wanted to rebuild the software for using the components to work with the API + client software ex. app or website panel, but I do not have enough time to do it. My repo can be pausey.

PR #382, opened 2018-05-26
Update autoupdate.sh

nik0:


jmtatsch:
Is that really new?
Imho this is already merged and I have made some rounds of spelling/grammar/etc changes to this...

nik0:
It is the previous version as now there is the "merge" tags in it:
line 6
`
<<<<<<< HEAD
# Depends on curl, jq (json parser), openssl (SHA calculation)
=======
# Depends on curl, jq (jason parser), openssl (SHA calculation)
>>>>>>> 571126e1e29240593c07125be9f83d21471dc9d0
`
It was my bad ...

jmtatsch:
Ah now I understand. But had already checked in the resolved version in the meantime.

PR #384, opened 2018-05-27
Update to custom_vendors.xml

theslackjaw:
Vendor Xiaomi already exists on tinycam so changed vendor name to avoid error. Changed model name to make it more meaningful. Merged in kszere's reboot command (previously wasn't working) and test sound commands.

### DIFF
--- a/integration/tinycam/custom_vendors.xml
+++ b/integration/tinycam/custom_vendors.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <vendors>
-    <vendor name="Xiaomi">
-        <model name="custom Application Layer" defaultPort="80" defaultRtspPort="8554">
+    <vendor name="XiaomiHack">
+        <model name="DaFang EliasKotlyar" defaultPort="80" defaultRtspPort="8554">
             <request name="Snap">/cgi-bin/currentpic.cgi</request>
             <request name="RTSP">/unicast</request>
             <request name="PtzMoveRelLeft">/cgi-bin/action.cgi?cmd=motor_left</request>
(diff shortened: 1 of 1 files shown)


integration/tinycam/custom_vendors.xml
```diff
@@ -1,15 +1,16 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <vendors>
-    <vendor name="Xiaomi">
```
jmtatsch: Really, there cannot be different models from one vendor?

integration/tinycam/custom_vendors.xml
```diff
@@ -1,15 +1,16 @@
...
+        <model name="DaFang EliasKotlyar" defaultPort="80" defaultRtspPort="8554">
             <request name="Snap">/cgi-bin/currentpic.cgi</request>
             <request name="RTSP">/unicast</request>
             <request name="PtzMoveRelLeft">/cgi-bin/action.cgi?cmd=motor_left</request>
             <request name="PtzMoveRelRight">/cgi-bin/action.cgi?cmd=motor_right</request>
             <request name="PtzMoveRelUp">/cgi-bin/action.cgi?cmd=motor_up</request>
             <request name="PtzMoveRelDown">/cgi-bin/action.cgi?cmd=motor_down</request>
             <request name="PtzMoveHome">/cgi-bin/action.cgi?action=motor_calibrate</request>
-            <request custom="Reboot" confirmation="yes">/cgi-bin/action.cgi?action=reboot</request>
+            <request custom="Reboot" confirmation="yes">/cgi-bin/api.cgi?action=reboot</request>
```
jmtatsch: Please use the original action.cgi whenever possible i.e. /cgi-bin/action.cgi?cmd=reboot
api.cgi contains a lot of duplicate code and is going to be deprecated eventually.

integration/tinycam/custom_vendors.xml
anemes: unfortunately not. Tinycam complains and doesn't read the file.

jmtatsch:
I have implemented some of your suggestions in the last commits. Please check if they work for you.

Issue #287, opened 2018-04-23
Support new DaFang revision without motor endstops


joshstrange:
I could't figure out why the web UI wasn't coming up (also ssh/ftp) so I telnet'd into in and was running the run.sh commands one by one. I got to the `motor hcalibrate` one and the device spun around till the power cable was at the front and then made a noise like it was trying to go further but couldn't. That command never would complete. I tried turning it off and adjusting the base of the unit to a number of different orientations but none worked. I commented out that line and now the device will start up without issue. Has anyone else had this issue?

jmtatsch:
I had almost the same issue once, the vertical motor was stuck to far up. While the camera was turned off, I moved it down a little. After that `motor vcalibrate` would run again and everything was fine.

jmtatsch:
I just set up another fresh dafang. There seems to be a new revision and the calibration doesn't seem to work anymore because the endstops are not found anymore (maybe different sensor supplier or rationalised away). The camera turns like 2 times, winds up the cable and is stuck. We should not use hcalibrate and vcalibrate for now as it could really damage our units.

jmtatsch:
Old Serial: DF3-00115712
New Serial: DF3-00344746

eekdood:
Just received my first Dafang and I'm seeing the same thing. My serial is within 10k of your "New Serial" and the calibration is broken. Also experiencing the same issues as #291 with regards to connecting to WiFi, even after commenting out the calibration steps. I can confirm that the camera works with the native app and is also booting to the SD.

EliasKotlyar:
Hello everyone,

I suppose that they have cheaped out at least one of the sensors per axis. The "old" Version of the Dafang used two sensors for detecting the limits, they were mounted on following pins:
#define MOTOR0_MIN_GPIO		77		
#define MOTOR0_MAX_GPIO		78	
#define MOTOR1_MIN_GPIO		41
#define MOTOR1_MAX_GPIO		60

Motor 0 = horizontal axis,
Motor 1 = vertical axis.

Can someone check how the GPIOs behave on the new Camera? I can then try to fix this issue by a different motor-driver. Remember to "uncomment" the motor-driver first, because you cannot use them while the motor-driver is loaded. You can rotate the camera "manually" to the end position and then check the pins.

QcVictor:
Same behavior for me with latest release, nothing work led orange led blink after motor start calibration even with motor hcalibrate and motor vcalibrate commented. 
[...]

PR #389, opened 2018-05-30
Add PTZ presets and other optimizations.

neroxps:
## Add PTZ preset function

I would like to set up the preset point method to control the PTZ motor. So I wrote PTZpresets.sh to calculated the target position with `motor -d s`.

 - **A firmware_mod/scripts/PTZpresets.sh**
 - **M firmware_mod/run.sh**
 - **M firmware_mod/www/cgi-bin/api.cgi**

**1. The motor calibration is fixed for new version of the device (please check whether the old DAFANG version device is working properly)**
**2. Add motor preset function (following are the method of usage)**

```bash
#/system/sdcard/scripts/PTZpresets.sh x_axis y_axis
/system/sdcard/scripts/PTZpresets.sh 1200 400
```

**3. Add PTZ presets API**

```bash
# http://username:password@DAFANG_IP/cgi-bin/api.cgi?action=motor_PTZ&x_axis=1700&y_axis=400
wget --output-document=/dev/null "http://username:password@DAFANG_IP/cgi-bin/api.cgi?action=motor_PTZ&x_axis=1700&y_axis=400"
```

## Add Display Axis Function

In order to set the presets of the pan and tilt for convenience, I added display axes to the OSD.

You can modify the `DISPLAY_AXIS=true | false` option in osd.conf to choose whether Axis is displayed.

 - **M firmware_mod/www/cgi-bin/action.cgi**
 - **M firmware_mod/www/cgi-bin/status.cgi**
 - **M firmware_mod/www/scripts/status.cgi.js**
 - **M firmware_mod/scripts/common_functions.sh**
 - **M firmware_mod/controlscripts/configureOsd**
 - **M firmware_mod/config/osd.conf**

## Add Limit Snapshots

The storage sapce will be filled up by unlimited Snapshot.

For this case,I would add the limit to the number of snapshots now.

You can adjust the maximum number of snapshots by setting the `max_snapshots=20` option in **motion.conf.**

 - **M firmware_mod/scripts/detectionOn.sh**
 - **M firmware_mod/config/motion.conf.dist**


jmtatsch:
motor calibration needs to work properly before we can merge this

firmware_mod/scripts/PTZpresets.sh
```diff
@@ -0,0 +1,119 @@
...
+echo $pid > /run/PTZ_$pid.pid
+
+# set log
+if [[ -z $3 ]]; then
+    LOG=false
+else
+    LOG=ture
+fi
+
+loger(){
```
jmtatsch: logger

firmware_mod/www/cgi-bin/api.cgi
```diff
@@ -224,6 +224,11 @@ if [ -n "$F_action" ]; then
      $BINPATH/motor -d h -s 100 &>/dev/null &
      getReturn 1234 "success" "Motor is calibration on horizontal."
   ;;
+  motor_PTZ)                                                                                                                                                            
+     /system/sdcard/scripts/PTZpresets.sh $F_x_axis $F_y_axis
```
jmtatsch: please don't use api.cgi anymore, it contains a lot of duplicate code and will be phased out soon

firmware_mod/scripts/PTZpresets.sh
neroxps: Sorry for my bad english.

firmware_mod/www/cgi-bin/api.cgi
neroxps: OK, I will move the code to action.cgi.

firmware_mod/scripts/PTZpresets.sh
```diff
@@ -0,0 +1,119 @@
...
+      return 1
+    elif [[ $2 -gt 800 -o $2 -lt 0 ]]; then
+      loger "Y should be between [0-800]"
+      return 1
+    fi
+    return 0
+}
+
+move(){
+	# Differentiate the coordinates and calculate the number
```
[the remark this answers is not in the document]
neroxps: The new version of DAFANG has removed the motor maximum step and minimum step sensor.
The current motor binary cannot be calibrated.
So I added the f option to motor binary for sensorless DAFANG to force calibration.
PTZ presets can only be used after calibration.
Do you have any other ways to solve the calibration? Or is there any better advice?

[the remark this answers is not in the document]
jmtatsch: If the version without sensor cannot be calibrated automatically, why did you add a fake calibration at all? 
We could kind of calibrate by hand and track the steps...

neroxps:
@jmtatsch 
The total number of steps at the motor level is 2500, and moving to the left is 0. If we move the motor to the left directly by the maximum number of steps (2500), make ensure that the motor is on the far left, then, the step count is cleared, that the horizontal calibration is completed. 

```C
        stepField_x = &status.x_steps;
        stepField_y = &status.y_steps;
        minField_x = &status.x_min;
        minField_y = &status.y_min;
        setMovement(MOTOR_DIRECTIONAL_LEFT, 2500);
        while (*stepField_x != -2500 && *minField_x != 1) {
            sendCommand(MOTOR_GET_STATUS, &status);
        }
        setMovement(MOTOR_DIRECTIONAL_DOWN, 800);
        while (*stepField_y != -800 && *minField_y != 1) {
            sendCommand(MOTOR_GET_STATUS, &status);
        }
        reset();
        setMovement(MOTOR_DIRECTIONAL_RIGHT, 1250);
        while (*stepField_x != 1250) {
            sendCommand(MOTOR_GET_STATUS, &status);
        }
        setMovement(MOTOR_DIRECTIONAL_UP, 400);
        while (*stepField_y != 400) {
            sendCommand(MOTOR_GET_STATUS, &status);
        }
        return 0;
```

neroxps:
>  It will wind up 1250 steps to the left on average per recalibration.

Turn 1250 left just to allow the camera to move to the middle position after calibration.
This does not affect the calibrated reference point.

But I found that the old DAFANG had a total vertical step size of 400, which was half that of the new DAFANG, so the camera rose to the maximum immediately and exceeded the step range after calibration, this is not the result I want.

> How about calling reset on the wake up point (0) of the camera. The user could then calibrate it by hand.

We can add the motor command with a z parameter such as `motor -d z`, this way can be set the current coordinate as the reference point.

> I think that is how it is handled for the newer generation in the original firmware.

I guess the official motor driver can detect the maximum and minimum step size of the new DAFANG motor by some methods.

@nik0 has also mentioned that there is currently a [(new) motor driver](https://github.com/beihuijie/Ingenic-kernel/tree/master/Ingenic_drivers/misc), but he has no time to adapt it.

I will not write C programs, and I only can modify the code by refer to previous code.

EliasKotlyar:
Hello everyone,

I have written a new application with the new motor driver. It was quite easy, and it should support also the new Dafangs. It uses the original kernel driver which is delivered with the dafang itself, and should have the same results as the original one. Can someone please test and report the results? 
https://github.com/EliasKotlyar/Xiaomi-Dafang-Hacks/issues/287

neroxps:
@EliasKotlyar 
No poblem, I will do my best.

firmware_mod/scripts/PTZpresets.sh
```diff
@@ -0,0 +1,104 @@
...
+
+MOTOR=/system/sdcard/bin/motor
+pid=$$
+echo $pid > /run/PTZ_$pid.pid
+
+# set log
+if [[ -z $3 ]]; then
+    LOG=false
+else
+    LOG=ture
```
jmtatsch: typo

neroxps: 😂
I have found this error.

I have found that there is still a problem with the motor command, so I have not yet pushed to github.

PR #390, opened 2018-05-30
Update rtsp-mjpeg

graystevens:
Any attempts to start the mjpeg service failed, as it was incorrectly calling 'stop' rather that 'start' on itself.


jmtatsch:
thank you for figuring this out.
nik0 seems to already have amended the path.

PR #416, opened 2018-06-05
Broken urls fixed in status.cgi

avegao:
This PR fix broken URLs in status.cgi


jmtatsch:
Hi, I don't understand why those urls seem broken for you. On the contrary, they were like absolute like you propose now a little while ago and we made them relative for people with reverse proxies.

PR #428, opened 2018-06-07
New auto updatepage

nik0:
- Add progress bar on autoupdate
- Add logs for auto update (on logs page)
- Use the last commit date on version field on settings page
- New v4l2server (sorry to add it on this commit) that solve the services that are not started when activated before the server start


jmtatsch:
looking forward to this 👍

PR #430, opened 2018-06-07
Adding missing jQuery and styles on update.html page

WatermelonFox:
I tried the update page, jQuery was missing and I also added the styles just in case.


WatermelonFox:
Indeed, I didn't take a look at that page!

PR #431, opened 2018-06-07
Adding missing jQuery and styles on update.html & logs.html 

WatermelonFox:
I tried the update page, jQuery was missing and I also added the styles just in case in the update.html page.
@nik0 also added the modification to the logs.html


nik0:
Thanks !

PR #439, opened 2018-06-08
Customize savedir

olskar:
Not sure if this is right at all, very new to this. I am trying to achieve the savedirectory for snapchats to be configurable in configuration instead of in the script


jmtatsch:
That should work. People will need to update their personal motion confs though.

PR #447, opened 2018-06-10
Add custom firmware for Wyzecam Pan

tssva:
Hardware wise the Wyzecam Pan is identical to the Dafang.


jmtatsch:
Very nice, thank you.

jmtatsch:
Do you know whether it has Endstop sensors or not?

jmtatsch:
"old" Version of the Dafang used two sensors for detecting the limits, they were mounted on following pins:

#define MOTOR0_MIN_GPIO	77
#define MOTOR0_MAX_GPIO	78
#define MOTOR1_MIN_GPIO	41
#define MOTOR1_MAX_GPIO	60

Motor 0 = horizontal axis,
Motor 1 = vertical axis.

Remember to comment the motor-driver first, because you cannot use them while the motor-driver is loaded. You can rotate the camera "manually" to the end position and then check the pins.

I used:

echo "41" > /sys/class/gpio/export
echo "60" > /sys/class/gpio/export
echo "77" > /sys/class/gpio/export
echo "78" > /sys/class/gpio/export
echo in > "/sys/class/gpio/gpio41/direction"
echo in > "/sys/class/gpio/gpio60/direction"
echo in > "/sys/class/gpio/gpio77/direction"
echo in > "/sys/class/gpio/gpio78/direction"
echo 1 > "/sys/class/gpio/gpio41/active_low"
echo 1 > "/sys/class/gpio/gpio60/active_low"
echo 1 > "/sys/class/gpio/gpio77/active_low"
echo 1 > "/sys/class/gpio/gpio78/active_low"

and then smth like
watch cat /sys/class/gpio/gpio78/value

tssva:
During my lunch hour I was able to ssh in, mount the debugfs and take a look at /sys/kernel/debug/gpio. Based upon it either there are no end stops or the motor driver that ships with it doesn't use them. Best bet is probably to assume they don't have any like the recent Xiaomi models. gpio 77 is being used for something but not an end stop. It is listed as sysfs which means a user space process has configured it. I'll have to look into that.

```
PIOs 0-31, GPIO A:
 gpio-18  (jxf22_reset         ) out hi
 gpio-25  (sysfs               ) out lo
 gpio-26  (sysfs               ) out lo

GPIOs 32-63, GPIO B:
 gpio-38  (sysfs               ) out hi
 gpio-39  (sysfs               ) out hi
 gpio-49  (sysfs               ) out lo
 gpio-51  (motor_st4_gpio      ) out lo
 gpio-52  (motor_st3_gpio      ) out lo
 gpio-53  (motor_st2_gpio      ) out lo
 gpio-54  (motor_st1_gpio      ) out lo
 gpio-57  (sda                 ) in  lo
 gpio-58  (scl                 ) in  lo
 gpio-59  (mmc_detect          ) in  lo
 gpio-62  (sdio_wifi_power_on  ) out hi
 gpio-63  (gpio_spk_en         ) in  lo

GPIOs 64-95, GPIO C:
 gpio-64  (home key            ) in  hi
 gpio-75  (motor_st1_gpio      ) out lo
 gpio-76  (motor_st2_gpio      ) out lo
 gpio-77  (sysfs               ) out hi
 gpio-79  (motor_st3_gpio      ) out lo
 gpio-80  (motor_st4_gpio      ) out lo

GPIOs 96-127, GPIO D:

GPIOs 128-159, GPIO E:

GPIOs 160-191, GPIO F:
```

jmtatsch:
Great analysis. So we can use gpio 77 to tell the different revisions apart? Or do you have an even better idea to do that?

tssva:
It appears both my Wyzecam Pan and Xiaomi dafang are lacking the end stops, so I can't test the behavior of units with endstops regarding the usb port. Is another gpio used to activate the usb port on them or are both the endstop and usb port wired to gpio 77 allowing only one or the other to be taken advantage of at any given moment. 

If it is the later than the best option is probably to install a motor driver without end stop support on all units, so the usb port can be taken advantage of.

PR #448, opened 2018-06-11
renamed lighttpd.conf & hostname.conf to .dist and init from it

drcobra:


jmtatsch:
possibly remove the conf files from the repo and add them to gitignore

drcobra:
I don't see the benefit to that. At least this way people can start with the default settings, and if they change them, the update function will not overwrite them.

PR #467, opened 2018-06-15
Check if the sdcard is mounted readonly and warn user

Dopeyr:
There's been a couple of instances where peoples sdcard have gone readonly and settings don't save. I've added a check for this and will warn them if it's not mounted read-write.


jmtatsch:
Thank you.

Issue #477, opened 2018-06-16
/usr/bin/basename doesn't work


Daviey:
basename normaly works because there is an alias in ~root/.busybox_aliases which means that the busybox in the PATH is used.

However, /usr/bin/basename is used (which is a symlink to /bin/busybox), it uses the older busybox which doesn't have the basename applet builtin.

```
[root@DAFANG:~]# /usr/bin/basename 
basename: applet not found
[root@DAFANG:~]# 
```

I've tried to think of some ways this can be solved, without touching the squashfs.. but we can't use symlinks, as this is a vfat filesystem.

The best i've been able to do so far is a bind mount:

```
mount -o bind /system/sdcard/bin/busybox /bin/busybox
```

nik0:
did you try `/system/sdcard/bin/busybox basename` ?

Daviey:
yeah, that works as expected.. but the issue is I want to use scripts which haven't been customised for this platform.

PR #478, opened 2018-06-16
bind mount /bin/busybox to sdcard one

Daviey:
Previously /bin/busybox was the squashfs shipped busybox binary, which
was lacking many features, even with some squashfs shipped symlinks such
as /usr/bin/basename (which is a compat symlink) not working.

This change bind mounts /system/sdcard/bin/busybox to /bin/busybox,
which effectively overrides the squashfs busybox.  This is done on each
boot via run.sh

Closes: #477


jmtatsch:
Thanks for contributing.

PR #479, opened 2018-06-16
Add standard periodic cronjobs

Daviey:
Previously adding a cronjob meant adding it to the root crontab with
appropriate syntax.  This change adds some standard periodic locations,
where simply dropping in a shell script into a folder means it will be
executed on schedule.  Adding specific jobs remains unaffected by this
change, and upgrades will not be touched.

The periodic jobs are:
  - 15min (every 15 mins)
  - hourly (on the hour)
  - daily (02:00 everyday)
  - weekly (03:00 on Saturday)
  - monthly (05:00 on first of month)

ie, adding a script to:
 - /system/sdcard/config/cron/periodic/15min/foo.sh
will execute every 15 mins.

### DIFF
--- a/firmware_mod/run.sh
+++ b/firmware_mod/run.sh
@@ -40,8 +40,24 @@ echo "Bind mounted /system/sdcard/etc to /etc" >> $LOGPATH
 
 ## Create crontab dir and start crond:
 if [ ! -d /system/sdcard/config/cron ]; then
-  mkdir -p /system/sdcard/config/cron/crontabs
-  echo "Created cron directory" >> $LOGPATH
+  mkdir -p ${CONFIGPATH}/cron/crontabs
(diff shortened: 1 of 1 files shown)


jmtatsch:
Thanks for your contribution.

lagunabravo:
It does not run scripts with .sh extension, does it?
I tried a script under the periodic cronjobs and it wouldn't run. I came across this https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=472585 and deleting the .sh extension in my script resolved the problem (ie. foo.sh to foo). 
@Daviey your example above does include the .sh extension (/system/sdcard/config/cron/periodic/15min/foo.sh). Can anyone replicate my problem?

numericOverflow:
I know it's a bit late/old, but I just ran into the same issue as @lagunabravo mentioned above and removing the ".sh" extension on my scripts makes it work flawlessly.

PR #483, opened 2018-06-17
Add Mozilla's CA Cert support to curl

Daviey:
Previously, curl wouldn't verify SSL certs when making outbound
connections.  It meant that it would either refuse to connect, or the
"-k" option needed to be added to acknowledge the insecurity and blindly
trust it. :/

This change adds the Mozilla CA store, repacked by upstream curl.  There
is also an update.sh script to update the pack if needed

As this is Mozilla's store simply repacked, it continues to be under the
MPL 2.0 license


jmtatsch:
Should we check for existence of cacert.pem in run.sh and download a fresh one if it doesn't exist?

Daviey:
Hmm.. I hadn't considered that.  I guess you mean as part of run.sh, as part of upgrades?

New installs will include the shipped one.

lex-obrien:
Before I open a a new separate issue, I wanted to make sure my thinking / process is correct :)  

I am building a Go app to run on the camera, and I was having trouble trying to connect to some services since Go is expecting the certs to be in /etc/ssl/certs.  After fighting with a bunch of things and disabling various items, I found this file and realized it was the certs I needed.  So I simply moved it over to the "normal" path of /etc/ssl/certs/[file here] and POOF! no more issues.

So, now my question. Is there a reason why we wouldn't want this file to be in that location to start with?  

Thanks in advance, and happy to take a stab at making a PR for this.

PS - please let me know if should have created this as a separate issue vs posting on said old PR.

jmtatsch:
Hmm, one of the design principles was not to touch the main filesystems except for the one link that runs run.sh if a sdcard is present. So if you take out the card everything is stock.
Of course this approach has also many limitations.
If you want a more sound linux layout and don't need all gui functions i would recommend looking into openfang firmwares.

PR #485, opened 2018-06-17
Update files copied to /system/sdcard/etc

tssva:
Each file needed from the original /etc is checked for on the sdcard individually and copied if missing. This allows adding to the list of needed files at a later date versus just checking for existence of the etc directory on the sdcard.

webrtc_profile.ini is added to the list of files copied to the sdcard. If automatic echo cancellation is ever enabled for audio this file is required.

A default passwd and shadow file are provided. Wyzecam has changed the default login password on the Wyzecam Pan. Providing default passwd and shadow files restores the default login password to ismart12. These files have been added to the exclude filter in autoupdate.sh.


jmtatsch:

Issue #491, opened 2018-06-17
Not able to change setting and to open "services" 


TheCrey:
Hello, 

Dafang

i do have an issue where i am not sure what exactly is the issue. 
I think a kind of full reset would solve it, bu I am not quite sure how to achieve this. 

I did have had all running and it was very fine, then I decided to run the cam with the originAl firmware, to see how it would behave with the ap etc. 
So, before all this happened, I was able to get into services to update the hack etc. 
So i removed just the sd and added it into the app.
Now I wanted to go back to the hacked version, but it was not working, so I repeated the install stell and this wen well. 

Now, I have access again, I also can see the view and a picture on the main page, but everything else it not working fine. 
As Example, If I want to change the name Dafang, I can save, but nothing happens, the name stays the same, also the movement does not work. 
The Stream does work to, even the settings say its turned off.
The Update area shows "update in progress" but nothing happens. 

The Logs are completely empty. Any of them. 

Network Information does not load also. 
SSH is working fine, but does not help me right now. 

While the Cam was connected to the APP, it installed an Update, to 5.5.1.327

Anyone can help?
Do you need any more information? 

Cheers

Damme:
I have same problem

hcgonzalezpr:
Same here. Status.CGI returns error 500.

nik0:
I guess PR https://github.com/EliasKotlyar/Xiaomi-Dafang-Hacks/pull/478 introduce the problem
Don't know why ...

hcgonzalezpr:
@nik0 if I enable logs on lighttpd does it give us the error?

One other thing that #478 might be affecting is root password, after last update, the root password returned to the default password.

nik0:
No,  no error
even strace does not give obvious error
I never change my root password

hcgonzalezpr:
@nik0  I manually reverted the changes made in #478 and can confirm that the gui works.

It seems the new busyxbox does not want to play nice.

Damme:
confirm solution works

tomusher:
My solution for this was to update all 

`source func.cgi`

to

`source /system/sdcard/www/cgi-bin/func.cgi`

in any relevant files under `/system/sdcard/www/cgi-bin`.

I've not figured out why the relative path stopped working with the mentioned change though.

Damme:
confirmed, solution works too :)

Daviey:
[...]

PR #492, opened 2018-06-17
Fix CGI execution and add cgi stderr logging

Daviey:
Change #478 introduced a regression with the updated busybox where ".",
seems not be part of the default $PATH.

The cgi scripts were sourcing "func.cgi", but this could not be found in
the new default $PATH, and therefore failed to execute. This meant that
the web-ui was unable to make changes and execute commands.

This has been resolved by putting a leading "./" for the script, making
it now ./func.cgi.

In addition, this also adds the lighttpd option server.breakagelog,
which enables stderr logging to /tmp/lighttpd-cgi-stderr.log (so 500
internal errors output is now logged to file (previously just the error
500 was in the webserver logs).

Closes: #491


nik0:
Thanks !

PR #493, opened 2018-06-17
Ugh, missile.cgi had a relative source aswell

Daviey:
In the previous commit, func.cgi sourcing was fixed by adding a leading
"./" to the filename.

However, I missed that missile.cgi had a cheeky source of header.cgi
which I missed in the previous change.

Closes: #491


Daviey:
@nik0 Sorry about missing that.. :/

PR #494, opened 2018-06-17
Close link to fix UI bug - entire page was "clickable"

wxmerkt:
The entire page was clickable and thus other settings could not be used as it always opened a new tab.


Dopeyr:
Woops! Thanks @wxmerkt

jmtatsch:
Thanks for catching that.

Issue #496, opened 2018-06-18
flash_eraseall: applet not found


patrik72:
tried to flash boot.bin with the actual release. no luck.
when i tried to erase mtd0 with this command: flash_eraseall /dev/mtd0, i received the message "applet not found". 
So how can i update the firmware, because i ran out of mem.

jmtatsch:
This is possibly caused by using the new busy box binaries in #478. flash_eraseall might be a custom applet of the old busybox.

jmtatsch:
Fixed in https://github.com/EliasKotlyar/Xiaomi-Dafang-Hacks/commit/75eee98412c560efbbad041d0abff0c37e2c73e4. Thanks @Daviey

timdonovanuk:
I just tried this and also got `flash_eraseall: applet not found `.  I'm currently using cfw-1.1.bin.

PR #497, opened 2018-06-19
Update busybox to include flash_eraseall

Daviey:
Previously flash_eraseall worked, because the vendor provided busybox
included the applet flash_eraseall.  However, the system busybox was
recently updated to point to the custom one.

This was missing the applet flash_eraseall, which this change fixes.  In
addition, this change updates the ~root/.busybox_aliases with the alias
(and verified there were no other missing applets).

Closes: #496


jmtatsch:
@Daviey thanks for the quick fix, but we have sort of a policy that we don't check in binaries not built by the core members. I added your improved compile recipe  https://github.com/Dafang-Hacks/Main/commit/0625d422bbaf3e69548ba54d7c48a078dcfd24b6, built it and committed it to the repository.

PR #500, opened 2018-06-20
Fix motion detection userscripts

Dopeyr:
There was a problem where only the first script was executed in `/system/sdcard/config/userscripts/motiondetection` - this should be fixed now.

I've also added a little documentation about usage of these scripts.


jmtatsch:
Thanks for your contribution.

PR #511, opened 2018-06-23
Only allow one instance of sendPictureMail to run at once

Dopeyr:
I found that because sendPictureMail was being run in the background that multiple instances could run at once. My concern was that for a reasonable amount of motion there are multiple copies of openssl/jpegoptim running, which started to slow my camera response.

I added a .lock file so only instance can run at once.


jmtatsch:
Thank you for contributing.

PR #513, opened 2018-06-23
Update mqtt-status-interval.sh

### LINKED ISSUE #288: Invalid payload for MQTT cover in HA
After upgrading I now get this issue with MQTT autodiscovery in HA:

> 2018-04-23 23:54:18 WARNING (MainThread) [homeassistant.components.cover.mqtt] Payload is not True, False, or integer (0-100): -86
> 2018-04-23 23:54:18 WARNING (MainThread) [homeassistant.components.cover.mqtt] Payload is not True, False, or integer (0-100): 172

Seems like the values for position needs to be between 0 and 100.

olskar:
Dont know if this is the best way, but "fixes" #288


jmtatsch:
This is rather fixing the symptom than the problem. Please try to fix the underlying problem that the output of the motor binary has changed...

jmtatsch:
Fixed via #537

Issue #236, opened 2018-04-16
Ethernet over USB adapter


vinceHH:
Hello,

I need to install the camera in a room where wifi reception is weak. However, this piece has an ethernet connection. Is it possible to connect the camera via a USB ethernet adapter?

thanks !

vinceHH:
hello,

how to add drivers to the firmware ?

thanks

jmtatsch:
Compile them with the toolchain in https://github.com/EliasKotlyar/Xiaomi-Dafang-Software
place them in the driver directory
ismod them in run.sh

vinceHH:
Hello jmtatsch,

thank you for your reply !
I'm happy to hear that it's possible to do it!
my only problem is that I understand the principle of using the toolchain to cross compiler drivers but it lacks some explanations to do it myself.
Could you describe a little more precisely the steps of using the toolchain?

do i have to use one of the compli _ *. sh as an example?

thanks !

vinceHH:
I try to compile the driver by myself but it seems that the makefile include with it used hardcoded path so, it's not working well :(

jmtatsch:
Then fix the hardcoded path?

vinceHH:
Hello,

Sure, it's my goal but it's beyond my skills because I do not know the path to use :(
For information, here the Makefile (renamed with txt extension) included with the driver.

[Makefile.txt](https://github.com/EliasKotlyar/Xiaomi-Dafang-Hacks/files/1991589/Makefile.txt)

is there a tuto to compile a driver with the toolchain ?

thanks!

jmtatsch:
Have a look at https://github.com/EliasKotlyar/Xiaomi-Dafang-Software/tree/master/drivers for other examples

vinceHH:
Thanks jmtatsch,

I make a try with an example from the drivers but I get an erro from gcc. In fact, I get the same error if I try to compile an existing driver :

```
make[1]: Entering directory '/tmp/Xiaomi-Dafang-Software-master/kernel'

  ERROR: Kernel configuration is invalid.
         include/generated/autoconf.h or include/config/auto.conf are missing.
         Run 'make oldconfig && make prepare' on kernel src to fix it.


  WARNING: Symbol version dump /tmp/Xiaomi-Dafang-Software-master/kernel/Module.symvers
           is missing; modules will have no dependencies and modversions.

[...]

PR #536, opened 2018-06-29
Enable support USB Ethernet

hcgonzalezpr:
Enable initial support USB Ethernet starting with the asix.ko driver.

This is my attempt to implement #236


jmtatsch:
Thank you very much. For security reasons we should not check in compiled code.
Would you be so kind to contribute your compile_ethernet_driver.sh recipe to https://github.com/Dafang-Hacks/Main ?

hcgonzalezpr:
@jmtatsch 
I don't know how to do that, I you can give me an example, im willing to give it a try.

But all I did was compile  https://github.com/Dafang-Hacks/kernel with the following entries as modules:
CONFIG_USB_USBNET=m
CONFIG_USB_NET_AX8817X=m

ulflun88:
Nice feature to have!

EliasKotlyar:
Many thanks. I checked the sources, it can be enabled trough the described kernel configurations. I enabled them by default in the kernel repository. So if someone is not trusting this modules, he can compile them by himself.

PR #537, opened 2018-06-30
HA cover value scaling

tomusher:
This is a fix for #288 which converts the internal motor positions to a 0-100 value required by HA.

This is done in the HA value template, so we can continue to use the full range of values elsewhere.

I've simplified the `motor status` in `common_functions` as it seemed not to be returning correct values, and the fixed endstop values don't seem to be being used anywhere.


jmtatsch:
Thank you for looking into this

PR #562, opened 2018-07-10
Fix video size setting in web ui Settings page

tookus123:
Video size gets stuck at 1600x900. Current grep isn't specific enough to pick up the video size setting from rtspserver.conf. It doesn't take into account  #commented lines with video sizes and by default, greps the "1600" part of AUDIOINBR=16000 which is why it's always reverts back to 1600x900 unless AUDIOINBR is set to a different value.  Since there's only 1 Video Size selector, both RTSPH264OPTS and RTSPMJPEGOPTS will always be set to the same size in rtspserver.conf. Therefor, grepping either of them should be sufficient to set the "selected" value.

**Maintainers:**  This is my first attempt at a pull request to please let me know if i did something wrong.

### DIFF
--- a/firmware_mod/www/cgi-bin/status.cgi
+++ b/firmware_mod/www/cgi-bin/status.cgi
@@ -391,11 +391,11 @@ cat << EOF
                         <div class="control">
                             <div class="select">
                                 <select name="video_size">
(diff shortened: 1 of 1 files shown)


jmtatsch:
Thank you for fixing those issues. Next time consider making a pull request for each issue fixed :)

tookus123:
You're welcome. Tried to make them separate but somehow they ended up under  the same commit. Git newbie mistake.

PR #579, opened 2018-07-17
Make uboot-flash script work from every location

Kaeltis:
flashinguboot.md tells the user to just SSH into the camera and execute "/system/sdcard/uboot-flash/flash_opensource_t20.sh" which did not work before (was looking for .bin files in the local folder).


jmtatsch:
Thanks

PR #597, opened 2018-07-24
Modify detectionOn script to use optimized jpeg instead

atiti:
This patch adds a new command which acquires and optimizes an image from
the camera. It also modifies the detectionOn.sh script to use the
optimized jpeg, thereby reducing the space need, and the wear on the
microsd card.

### DIFF
--- a/firmware_mod/bin/getoptimimage
+++ b/firmware_mod/bin/getoptimimage
@@ -1,0 +1,13 @@
+#!/bin/bash
+#
+# The following script acquires an image from the camera into ramfs (/tmp)
+# Then optimizes the output using the jpegoptim command, and writes it to the
+# destination folder, thereby reducing space requirement, and wear on the microsd card
+#
+
+targetdir="$1"
+targetfile="$2"
+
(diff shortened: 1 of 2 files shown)


saipsa:
It should be optional based on configuration parameter.

nik0:
`--stdin` option ?
Or use jpegtran
`/system/sdcard/bin/getimage |  /system/sdcard/bin/jpegtran -progressive -optimize`

M0ebiu5:
It only works here, if i change the shell:
#!/bin/bash -> #!/bin/sh

Dopeyr:
@atiti You may be able to close this now, as in the latest firmware `/system/sdcard/bin/getimage` returns a much smaller, already compressed image.

e.g. 

Old firmware
---

```
MBP~ % ssh root@192.169.1.221 '/system/sdcard/bin/getimage|wc -c'
1440000
```

New firmware
---

```
MBP~ % ssh root@192.169.1.220 '/system/sdcard/bin/getimage|wc -c'
171167
```

So it's now roughly 10% the size.

atiti:
Great, then this is not needed :)

PR #598, opened 2018-07-24
Provide the captured image as a param to userscripts on motion detection

atiti:
This patch allows the user scripts to act directly on the captured image
by passing the path to the scripts directly as the second argument.
This way it's possible to work with the image directly, without
employing dirty tricks to find out which one was the latest image in the
stills directory.


saipsa:
It will work only if save_snapshot is true, otherwise path will point to no file, It should not be passed in those cases.

Dopeyr:
Yes, that sounds fair to me @atiti - the other scripts run in the background too. Depending on your use case you may want to add some locking inside the user script, so it doesn't run concurrently when there are many motion events.

PR #602, opened 2018-07-27
add missing port to homekit example

timoschilling:


jmtatsch:
That port should already be the default. It doesn’t work for you without?

PR #613, opened 2018-08-03
When sending mail, allow option for 'default' compression which allows for shorter gaps between images

Dopeyr:
Compression seems to be approx. 4x as fast than when specifying the -m option, but the still approx. 90% smaller than the original file size


nik0:
Did you test with the last server ? The generated jpg should be smaller

Dopeyr:
No, I'm quite a bit behind. I'll try it out and report back. Thanks!

Dopeyr:
Hi, yes that works great, looking about 100k per file (1600x900), so to speed up the send email time, it's just a case of setting QUALITY to -1 so jpegoptim doesn't run. Thanks :)

PR #628, opened 2018-08-08
Add MQTT send mail alert switch

gysmo38:
I try to add MQTT send mail alert switch.
It work for me. 
I add it because I would like to use motion like motion tracker for the house all the time and have mail alert only when away.

### DIFF
--- a/firmware_mod/scripts/common_functions.sh
+++ b/firmware_mod/scripts/common_functions.sh
@@ -290,6 +290,30 @@ motion_detection(){
   esac
 }
 
+# Control the motion detection mail function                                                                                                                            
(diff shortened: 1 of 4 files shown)


gysmo38:
You want another name?
It is a switch to activate or deactivate sending mail on motion detection.

jmtatsch:
for on you listen on /motion/send_mail/set
for off you listen on /motion/detection_mail/set
should probably both be /motion/send_mail/set?

gysmo38:
You are right :)  Thank you.

jmtatsch:
Looks good to me

PR #631, opened 2018-08-10
/system/sdcard/run.sh: make sure ntpd gets run AFTER wifi and internet is working (with timeout)

reikred:
Change
/system/sdcard/bin/busybox ntpd -p "$ntp_srv"
to 
timeout -t 30 sh -c "until ping -c1 "$ntp_srv" &>/dev/null; do sleep 3; done";
/system/sdcard/bin/busybox ntpd -p "$ntp_srv"

This will wait up to 30sec for wifi and internet and ntp timeserver access to start working.
If this is not done, a race condition often causes time to be set to 1970 (unix epoch start),
and time may never get updated.


nik0:
Thanks !

PR #647, opened 2018-08-20
Remove oldest snaps when save_dir is full & run user motion scripts in background

saipsa:
The removal is done in background, as it blocks for 2 or 3 seconds.


saipsa:
Yo, someone merge it. At present the removal is done in the default order, which is by name so newest snaps were removed instead of the oldest.

jmtatsch:
Thank you for catching that

PR #658, opened 2018-08-26
Fix grep_text for correct output from bin/motor

### LINKED ISSUE #656: motor_PTZ only goes right and up
Not sure what im doing wrong but the parameters to motor_PTZ seem to increment :)
Ex:
cmd=motor_PTZ&x_axis=100&y_axis=100
Will move 100 on each axes...

nathanchantrell:
Output from bin/motor has changed, updated to reflect this. Fixes #656

### DIFF
--- a/firmware_mod/scripts/PTZpresets.sh
+++ b/firmware_mod/scripts/PTZpresets.sh
@@ -41,11 +41,11 @@ move(){
 	# Differentiate the coordinates and calculate the number
 	# of steps to reach the specified coordinates.
 	if [[ "$1" = "X" ]];then 
-		grep_text="x_steps"
+		grep_text="x"
 		opt="r|l"
 		text="Right|Left"
 	else
-		grep_text="y_steps"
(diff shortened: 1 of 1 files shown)


nik0:
Thanks catching this ...

pixeldublu:
That was fast :) 

Thumbs up!

pixeldublu:
Still not ok... now its inverted... always goes left :)

pixeldublu:
Maybe string lower of $1 and compare it a gains "x"

nathanchantrell:
Oh that's odd. It's working correctly for me with just that change and I can't see anything else that would cause that.

If you log in via SSH and run /system/sdcard/bin/motor -d s
do the x and y values it returns increment and decrement as you move the camera?

pixeldublu:
Will test tonight when i get home.

I know that motor_left motor_right works correctly.

pixeldublu:
I think its ok now. Thanks alot!!

PR #666, opened 2018-09-01
Fix for missing port parameter in mqtt-autodiscovery.sh

TranscendOfSypherus:
Fixed missing port parameter when publishing motion detection snapshot detail to MQTT


jmtatsch:
Thank you for contributing.

Issue #675, opened 2018-09-04
mqtt-control reads the messages it sends


jboeddeker:
mqtt-control.sh listens to $TOPIC/# while it sends messages the same topic. So, whenever a control message is received, the just-sent answer is also received and read. 

According to my opinion, it would be better configure the server to listen to a special control topic $TOPIC/control/#, then it could ignore all the status messages it is sending by itself. 

I am able to change this, but i do not know if that separation of in/out queue is possible with all smarthome systems. I only use openHAB with that it would be fine.

Another option would be to include all known command topics to the mosquitto_sub -t parameter, but this would be quite a list and not really maintainable.

jmtatsch:
First option seems alright. Can you whip up a PR?

jboeddeker:
Yes. I will use the $TOPIC/control.

jboeddeker:
The problem is the performance. Every status message and every answer triggers the switch statement with its 30 something cases and causes another invocation of mosquitto_sub. If you send several commands you will notice a delay.

Add an echo $line in the control loop and execute mqtt-control.sh in a shell. Nothing wrong is happening, just too much unnecessary stuff.

jmtatsch:
I understand your concerns now and agree.
But when we are changing the api at all I would like to take the opportunity to make it more simple and consistent. I will make a pull request soon and request for feedback.

guandalf:
Hi,

I am available to give it a try on Home Assistant, if this can help.

TY

jmtatsch:
@guandalf that would be swell.

guandalf:
I applied the patch but I get:

[root@DAFANG:sdcard]# /system/sdcard/bin/busybox /system/sdcard/scripts/mqtt-control.sh
mqtt-control.sh: applet not found

running mqtt-control I had no time to investigate yet.
Trying tonight.

jboeddeker:
@guandalf, could you tell me how you did apply the the patch? Thanks.

guandalf:
I downloaded the patch from the pull request:
wget https://patch-diff.githubusercontent.com/raw/EliasKotlyar/Xiaomi-Dafang-Hacks/pull/677.patch

I removed some junk with vi, then:
patch -p1 < 677.patch
on a local copy of the repo and I "uploaded" the patched files pasting them with vi.

jmtatsch:
@guandalf missing a nohup?

jboeddeker:
[...]

PR #677, opened 2018-09-04
Add "control" topic to keep mqtt-control from reading and parsing its own messages

jboeddeker:
mqtt-control.sh listens to $TOPIC/# while it sends messages the same topic. So, whenever a control message is received, the just-sent answer is also received and read. #675

According to my opinion, it would be better configure the server to listen to a special control topic $TOPIC/control/#, then it could ignore all the status messages it is sending by itself.


jboeddeker:
#675

PR #678, opened 2018-09-05
Fixing telegram integration in detectionOn.sh

gergles:
The (+) symbols on lines 48/49 of the original file look like they were left there from some other automated process or diff but they break the Telegram integration of the script.

Additionally, if you don't have save_snapshot set to true, the cwd that this script runs in does not work for saving images, so getimage fails.  To fix both these problems, I removed the +es and added in the $save_dir reference.

Additionally, the block that creates the $save_dir if it doesn't exist should probably be replicated too down in the Telegram section if $save_snapshot is false, but I didn't put it there because I'm not sure if you might just want to have that run one time in common_functions or something like that.


jmtatsch:
Thanks for catching that

PR #752, opened 2018-10-23
Updating autonight software method

chisht:
1. Integrating sw_night_configure.html to main index.html
2. Adding GUI option to select SW/HW mode for autonight in setting page
3. Adding capability to read current configuation and save the new configuation back to the sdcard in sw_night_configure

### DIFF
--- a/firmware_mod/www/cgi-bin/action.cgi
+++ b/firmware_mod/www/cgi-bin/action.cgi
@@ -447,6 +447,32 @@ if [ -n "$F_cmd" ]; then
         echo "Motion Configuration done"
         return
     ;;
+    autonight_sw)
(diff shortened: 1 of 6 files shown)


chisht:
Sure. My concern with removing autonight.conf.sw is that if we remove the file then hitting the radio button to select HW mode for autonight, will remove the conf. And if you had some custom settings in there, then it'd be gone. If you are ok with that limitation, then sure, I can make the change.

chisht:
Actually all we need to do to switch back and forth between software and hardware mode of autonight is to add or remove -S option. In hardware mode, rest of the options don't matter. And code will ignore it.

I'll work on changing this PR to do just that.

PR #771, opened 2018-11-02
introduce telgram bot service

davydes:
Simple telegram bot.

Opportunities:
  - take a shot /shot
  - free memory info & swap usage /mem
  - on/off motion detection


davydes:
i need remaster it

Issue #754, opened 2018-10-24
Updating from webinterface is broken


saipsa:
Updating from webinterface replaces all the files with 0 byte files. I tried it on two different sd cards.

Dopeyr:
There should be output in /system/sdcard/log/update.log - does that reveal any issues?

saipsa:
Here is the log https://pastebin.com/raw/ERD8YGk1

It seems, all files were forced updated in the last update and it resulted in 0 byte files.

Dopeyr:
Roughly when did you update? Just wondering if it's something to do with the github outage around 22nd/23rd October? See https://status.github.com/messages Just wondering if the files were downloaded as 0 bytes because of the outage.

Is it still happening now? 

For what it's worth, the same problem was likely to occur if you ran the update via SSH as it uses the same script underneath, so this issue isn't just restricted to the web interface.

Regardless, perhaps the autoupdate script can be modified to detect 0 byte files and produce an error.

saipsa:
The problem happened a few weeks ago. I ignored it, attributing it to chance error.I reformatted the card and copied the files from the repo. 


I updated it yesterday and camera failed to boot again, so I thought it is not due to chance error. 

Maybe resource throttling by github based on my ISP? 

Nevertheless, zero byte check should be added in the update script.

jmtatsch:
Would be happy to accept a PR

saipsa:
Sure. I will create it in a day or two.

columbusgeek:
This happened to me last night too after pressing update and walking away for the night.  I thought I just had a dropped connection or something.  When I eyeballed the sdcard today I noticed the 0 byte files.   
Easy enough fix, just thought you'd like to know it happened to me as well.

saipsa:
@jmtatsch I have created a PR

PR #783, opened 2018-11-09
Fix for zero byte files overwriting local files

saipsa:
This commit fixes zero byte files overwriting local files, making camera unbootable

Fixes https://github.com/EliasKotlyar/Xiaomi-Dafang-Hacks/issues/754

### DIFF
--- a/firmware_mod/autoupdate.sh
+++ b/firmware_mod/autoupdate.sh
@@ -260,6 +260,12 @@ do
         echo "Can not get remote file $i, exiting."
         exit 1
     fi
+    # sometimes zero byte files are received, which overwrite the local files, we ignore those files
+    # exception: files that are hidden i.e. start with dot. Ex: files like ".gitkeep"
+    if [[ ! -s ${TMPFILE} ]] && [[ $(basename ${LOCALFILE} | cut -c1-1) != "." ]
+        echo "Received zero byte file $i, exiting."                             
+        exit 1                                                                  
+    fi   
     # Check the file exists in local
     if [ -f "${DESTFOLDER}/${LOCALFILE}" ]; then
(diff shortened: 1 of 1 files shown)


jmtatsch:
Thanks

cwiggs:
So I tried to update today and I believe I hit the exit that was introduced by this PR.

The autoupdate.sh script exits but doesn't explain what needs to be done to fix the issue.

How do I go about updating now that the autoupdate.sh script just keeps exiting?

cwiggs:
Just an update on this.  I added an `ls -l` to the `{TMPFILE}` var to see which file is zero bytes and it comes back with this file:

`-rw-r--r--    1 root     root             0 Mar 23 08:25 /tmp/update.tmp`

cwiggs:
I think I found the actual issue:

```
# df -h | grep /tmp
tmpfs                    48.6M     48.6M         0 100% /tmp
```

```
# ls -l /tmp
total 49768
-rw-r--r--    1 root     root            32 Feb 17 12:18 isp_tuning_func
-rw-r--r--    1 root     root      50946048 Mar 21 21:11 lighttpd-access.log
-rw-r--r--    1 root     root           194 Mar 21 20:34 lighttpd-cgi-stderr.log
-rw-r--r--    1 root     root          4096 Mar 21 21:11 lighttpd-error.log
-rw-r--r--    1 root     root             0 Mar 21 20:34 progress
-rw-r--r--    1 root     root             0 Mar 21 20:33 update.log
-rw-r--r--    1 root     root             1 Mar 22 10:13 wpastat
```

/tmp was full because of the lighttpd-access.log file being so large.

I have home assistant grabbing images every few seconds from DAFANG, so there is a line in log that is showing 200 from home assistant every few seconds.  Looks like we might want to add some type of logrotate for that file?

Anyway, this PR wasn't the actual issue, in fact I think it saved me from doing something bad to the device; Thanks!

PR #786, opened 2018-11-10
Use common_function for action.cgi's motor calls

### LINKED ISSUE #784: OSD (Axis position) not updated on mqtt PTZ
Xiaomi Dafang Camera, running 9a664c9

### Description

After sending motion commands via mqtt, the camera moves, but the OSD is not updated.

Moving the camera using the default web UI does update the OSD

### What did you do to debug the issue

Comparing the code in mqtt-control.sh, it seems like it should call
```bash
      update_axis
      /system/sdcard/bin/setconf -k o -v "$OSD"
```
possibly after a sleep, after each motion.

Perhaps a better place would be in `common_function.sh`. What do you think? I'm happy to submit a PR, but I'd like to first understand if there's a reason this was done in the web UI, not in common_functions, and if there's any reason not to do it for mqtt (blocking, perhaps?)

mikeage:
Closes #784

### DIFF
--- a/firmware_mod/scripts/common_functions.sh
+++ b/firmware_mod/scripts/common_functions.sh
@@ -157,24 +157,31 @@ motor(){
   case "$1" in
   up)
     /system/sdcard/bin/motor -d u -s "$steps"
+    update_motor_pos $steps
     ;;
   down)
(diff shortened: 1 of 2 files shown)


jmtatsch:
Thanks for helping.

PR #791, opened 2018-11-13
Add a new parameter 'publish_mqtt_snapshot'

mikeage:
Currently, publish_mqtt_message is used to decide whether we send a
notification about motion. save_snapshot is used to save the image
locally. And if both are set, the image is also sent to mqtt

This PR decouples the mqtt image send from the snapshot saving. There
are now three parameters:

publish_mqtt_message -- send notification that there was motion
publish_mqtt_snapshot -- send notification containing the motion.
save_snapshot -- saves to disk, doesn't necessarily notify

No MQTT topics are changed; this is purely configuration bookkeeping.

### DIFF
--- a/firmware_mod/config/motion.conf.dist
+++ b/firmware_mod/config/motion.conf.dist
@@ -13,9 +13,13 @@ motion_tracking=off
 # Configure how to deal with motion events:
 
 motion_trigger_led=true
-save_snapshot=false
 max_snapshots=20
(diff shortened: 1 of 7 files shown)


mikeage:
note: this is not yet tested. I'll remove the `[WIP]` tag when I've finished my work

jmtatsch:
Thank you for contributing.

Issue #787, opened 2018-11-11
mqtt cover for left/right & up/down not valid in the latest homeassistant (0.82)


mikeage:
After updating to homeassistant 0.82, the covers for a dafang device (as discovered by the autodiscovery) no longer show a slider and the logs complain about an invalid value:
```
Nov 11 12:26:49 raspberrypi hass[22815]: 2018-11-11 12:26:49 WARNING (MainThread) [homeassistant.components.cover.mqtt] Payload is not True or False: 0
Nov 11 12:26:49 raspberrypi hass[22815]: 2018-11-11 12:26:49 WARNING (MainThread) [homeassistant.components.cover.mqtt] Payload is not True or False: 41
```
In #288 , the scaling was addressed, but it seems like it now only access a boolean. I don't know enough about mqtt cover types to know if this is valid, or we need to somehow tell it to get a boolean or integer.

mikeage:
The bug on HA's side is https://github.com/home-assistant/home-assistant/issues/18354 . The analysis there, while obvious, suggests that values are not valid; only boolean.

[note that the 0 in my log is not False, but simply the bottommost position; the 41 is left/right]

jmtatsch:
We need to use get/set_position_topic for the  numerical values and the state topic for the boolean.

dgomes:
HA changed mqtt_cover configuration (there is the issue that online docs have not been updated accordingly)

we must now use **position_topic** instead of **state_topic** for cover's

mikeage:
I made this change locally, and I see it's sending `position_topic` when I look at mosquitto_sub, but I'm still getting the errors in HA. Is there somewhere else that needs to be changed?
```diff
diff --git i/firmware_mod/scripts/mqtt-autodiscovery.sh w/firmware_mod/scripts/mqtt-autodiscovery.sh
index 43147ec..68cea94 100644
--- i/firmware_mod/scripts/mqtt-autodiscovery.sh
+++ w/firmware_mod/scripts/mqtt-autodiscovery.sh
@@ -47,5 +47,5 @@ VER="Dafang Hacks"


 # Motor up/down/left/right
[...]

PR #796, opened 2018-11-14
Fix cover definition in mqtt autodiscovery for HA 0.82+

mikeage:


mikeage:
Closes #787

jmtatsch:
Does this already work?

jmtatsch:
Thanks for fixing this.

badabing2005:
Please see [here](https://github.com/home-assistant/home-assistant/issues/18354)
As it seems to still be problematic.

Thanks
BB

PR #833, opened 2018-12-12
up/down note

gdelory:


gdelory:
I just noticed that when trying to integrate to nodered, and I thought I would put it so people don't waste time finding it.

jmtatsch:
Fixed in differd7ad6d5. Thank you for catching this.

gdelory:
sure anytime, your hack is simply awesome, I hope I would be able to help more in the future :)

PR #844, opened 2018-12-19
Improve Telegram bot resistance to temporary network or server issues

liads:
This should prevent the script from exiting unless there's an authentication error.
Also added logging of those occasional server errors.


jmtatsch:
Thank you for contributing

PR #853, opened 2018-12-27
Update telegram-bot-daemon.sh

blakadder:
Added /help command to the bot just in case someone forgets all the commands


jmtatsch:
Thanks.

PR #855, opened 2018-12-28
Telegram bot: Option to configure alert type - text/image

saipsa:


firmware_mod/scripts/common_functions.sh
```diff
@@ -3,6 +3,10 @@
 # This file is supposed to bundle some frequently used functions
 # so they can be easily improved in one place and be reused all over the place
 
+include () {
+    [[ -f "$1" ]] && source "$1"
```
jmtatsch: where is the advantage of this way to include?

firmware_mod/scripts/detectionOn.sh
```diff
@@ -4,6 +4,8 @@
 . /system/sdcard/config/motion.conf
 . /system/sdcard/scripts/common_functions.sh
 
+include /system/sdcard/config/telegram.conf
```
saipsa: @jmtatsch here is the benefit of including this way. Users may not have renamed `telegram.conf.dist` to `telegram.conf`. `detectionOn.sh` will fail in that case.

jmtatsch:
Looks good to me, thank you for helping.

PR #866, opened 2018-12-30
Added Telegram functionality to HA and webui

blakadder:
- added switch to activate or deactivate sending snapshots via Telegram
- added toggle for sendin snapshots via Telegram

Added the functionality cause I like the mqtt snapshot feature in HA but didn't like all the Telegram notifications if its enabled in config


blakadder:
Tested on my dafang camera, HA autodiscovery works, command line works, webui works (altho i just tested it a couple of times)

jmtatsch:
Thank you.

PR #867, opened 2018-12-31
Add motor variables and mqtt-control for position

alandtse:
This adds a set channel for MQTT motor positions and uses variables to set the min, max, and step based off motor settings from [here](https://github.com/EliasKotlyar/Xiaomi-Dafang-Hacks/blob/master/hacks/drivers.md).


jmtatsch:
Thank you for contributing.

Issue #652, opened 2018-08-22
Xiomi Dafang motion tracking always goes to bottom left


CSmither:
Everything installed correctly no obvious errors in logs but only just installed today so not really sure what I am looking for.

The motion tracking seems to work in so far as it detects movement and moves the camera however the camera always moves to the bottom left, I can move it with the arrows on the web interface but it then returns back to the bottom left when I stop instructing it.

Resolution set to 1280
video format set to vbr
stream works fine

Any other info needed just ask :+1:

akinomeroglu:
Hey all,

I suffer from this bug too.

nik0:
Do you have some server logs? To see if the problem is on the detection side or on the moving script

akinomeroglu:
Hey @nik0 could you kindly tell me where the logs stored. I can share it tonight.

Thanks,

nik0:
On the web interface go to "informations/logs" and "video server" tab
You should have something like this:
`Will execute command /system/sdcard/scripts/detectionTracking.sh 1 0 0 0`

If you can ssh to the camera you can run theses commands:
`/system/sdcard/scripts/detectionTracking.sh 1 0 0 0`
`/system/sdcard/scripts/detectionTracking.sh 0 1 0 0`
`/system/sdcard/scripts/detectionTracking.sh 0 0 1 0`
`/system/sdcard/scripts/detectionTracking.sh 0 0 0 1`

akinomeroglu:
@nik0 hello again sir,

Sorry for delay. Please let me know if you need something else.

`[root@DAFANG:~]# /system/sdcard/scripts/detectionTracking.sh 1 0 0 0
1 0
0 0
Left move
x: 0
y: 0
speed: 900
Up move
x: 0
y: 100
speed: 900
[root@DAFANG:~]# /system/sdcard/scripts/detectionTracking.sh 0 1 0 0
0 1
0 0
Right move
x: 100
y: 100
speed: 900
Up move
x: 100
y: 200
speed: 900
[root@DAFANG:~]# /system/sdcard/scripts/detectionTracking.sh 0 0 1 0
0 0
1 0
Left move
x: 0
y: 0
speed: 900
Down move
x: 0
y: 0
speed: 900
[root@DAFANG:~]# /system/sdcard/scripts/detectionTracking.sh 0 0 0 1
0 0
0 1
Right move
x: 100
y: 0
speed: 900
Down move
x: 100
y: 0
[...]

Issue #567, opened 2018-07-12
Motion track not really working


pixeldublu:
Hello,
New here and got a fresh installed dafang camera.
Wanted to activate the motion track but as soon as it senses some motion it rotates backwards and points camera down. 
Fresh and latest install few hours ago.
Is this a known issue ? missing some config ? 

Thanks in advance.

nik0:
What does the log of the video server says ?

staudtm:
Same with my camera. I have mounted my camera top down and thought this is the issue...

SiliconAvatar:
I resubmitted this as [issue #802](https://github.com/EliasKotlyar/Xiaomi-Dafang-Hacks/issues/802) with some new information.

To fix this, temporarily, First you have to change "detectionTracking.sh". You can do this from reading the SDCard externally, or use VI inside SSH; comment out or delete lines 103,104, and 105. This will stop the camer from overwriting the home position. next, SSH into the camera, go to the /tmp/ directory, and run "touch cameraposition" this will create the file that holds the "home" position. Then edit it in VI and type in X: xxxx" and "Y: xxxx: where the 'xxxx' sections hold the coordinates you want the camera to return to. Add a line break between them so the file is 2 lines long, one has the X position and the other the Y. You can find the position you want it to return to by first running the calibration, then centering the camera where you want and either looking at the position on the OSD, or running "/system/sdcard/bin/motor -d s" in the camera SSD command prompt. 

Once you have "detectionTracking.sh" no longer overwriting the position, and have the position file created and entered in the home position, it should work fine. Until it's rebooted, then the tmp directory is deleted. If you want to make this persistent, change line 16 of "detectionTracking.sh" to somewhere inside "/system/sdcard" and create the file there which will hold the home position of the camera.

Hopefully this issue get's resolved the proper way though, so this isnt needed.

PR #868, opened 2018-12-31
 Fix return to 0,0 origin for detectionTracking

### LINKED ISSUE #802: Motion Tracking Issue (always returns to 0,0 and not last position)
I have an issue with my WyseCam Pan camera running Dafang-Hacks, where when motion tracking is enabled, and the "Motion Timeout" is set to a positive number (IE 15) the camera always returns to the bottom left position (0,0) instead of the last known position as it should.

The Motion Timeout setting is described as "Restore camera position after x seconds, -1 to deactivate" which I take to mean that it will track motion, then after there is no more motion, it will return to it's original coordinates. It does not currently do that.

I think I've narrowed down the issue a bit.

There seems to be at least 3 separate issues in play here as far as I can tell, all seem to be in detectionTracking.sh.

First: The script detectionTracking.sh is meant to store the camera's original position to /tmp/cameraposition (line 16) however it never creates this file if it does not exist. I've tried this multiple times while SSH'ed into it watching that folder, and the file "cameraposition" never gets created. If I run "touch cameraposition" inside the /tmp/ folder, the script then starts storing the coordinates in that file as it should. There the logic to create the file if it does not already exist should be revised.  

Second: The script seems to store the camera's position every time there is no motion detected (Line 103). This is a problem. For example, if I have the motion timeout set to 15 seconds, the intention would be for the camera to sit at the "home" position, wait for motion, then track said motion. After no more motion after 15 seconds, it returns to 'home'. The problem is that every time the scripts stops seeing motion, it over-writes the 'home' position with where it was last at when the motion stopped. This will lead to the home position gradually walking away from where it was originally set to. I would think that this could be fixed by only writing the original position to the file at the start of a motion event, then not overwriting it until the entire motion event is over (taking the timeout into consideration) and already returned to base. Alternatively, the home position could be set once, by the user, then never overwritten.

Third: The script gets the camera's position by calling "motor -d s" (Line 104). However if the camera is currently traveling when this is called, it will return 0 for both the X and Y axis. There needs to be a sleep in here or something so that this is only called once the camera is stationary. 

I'm not enough of a programmer to know how to fix these, but I think my observations were correct. If I got something wrong, please let me know.

This is also likely a duplicate of Issue#567 and Issue#652, so it seems like a problem at least a few users are seeing.

alandtse:
Resolves #802, #652, #567.

Specifically uses analysis from SiliconAvatar in #802 to:

- Create /tmp/cameraposition on first run only.
- Leverage script [PTZpresents.sh](https://github.com/EliasKotlyar/Xiaomi-Dafang-Hacks/blob/master/firmware_mod/scripts/PTZpresets.sh) since it creates a pid lock to indicate motor action and handles x and y movement.  This probably should be used universally for motor control.
- This is based on my dev branch which includes pending PR #867


jmtatsch:
Thank you for contributing.

PR #882, opened 2019-01-12
Add a MQTT config variable to disable communicating the LDR status

nbarrientos:
In view of the devices that now come with no hardware light sensor and to save some traffic in the message queue, add a configuration variable to be consumed by `mqtt-status-interval`, `mqtt-control` and `mqtt-autodiscovery` to select if the sensor is published and updated.


jmtatsch:
Thank you for helping to improve this project.

PR #892, opened 2019-01-23
Streamline U-Boot installation instructions

akutruff:
Shows how to safely use wget and removes path ambiguities.


akutruff:
Is that better?  I'm glad you want to make this as safe as possible.

akutruff:
Also, I created a second pull request https://github.com/Dafang-Hacks/uboot/pull/3 to update the uboot probject

jmtatsch:
Thanks for clarifying the process.

PR #898, opened 2019-01-29
Simplify tracking algorithm

tsvi:
Instead of creating a huge if/elif, with an elif for every possible case, create 4 separate if statements.

### DIFF
--- a/firmware_mod/scripts/detectionTracking.sh
+++ b/firmware_mod/scripts/detectionTracking.sh
@@ -84,40 +84,23 @@ fi
 # Basic algorithm to calculate the movement
 # Not optimized, if you have ideas to simplify it ...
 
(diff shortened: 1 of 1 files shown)


firmware_mod/scripts/detectionTracking.sh
```diff
@@ -84,40 +84,23 @@ fi
 # Basic algorithm to calculate the movement
 # Not optimized, if you have ideas to simplify it ...
```
mikeage: Perhaps this comment can now be changed ;-)

tsvi: It could maybe be optimized better

jmtatsch:
Thanks for helping.

Issue #880, opened 2019-01-10
SD card turns to Read-Only everytime power is cycled


kaymach:
Hello. This is pretty consistent and 100% reproduce-able. Anytime I lose power to the camera or disconnect the power cable, the SD card will turn to read-only and I must plug it back into my PC to let windows fix the drive. Once I re-seat it back into the camera, it works fine again until the next power cycle down the line.

Installing it outdoors was a nightmare as I had to constantly take apart the outdoor enclosure and climb up a ladder to fix the SD card as I was not able to SSH in to remount it. It refused the connection.

The last time this issue was posted, it closed as the user just basically did what I did. To avoid future users and myself the headache of having to run around in case of power cycle, it would be useful to know how to permanently fix this issue going forward.

Any ideas?

MiguelAngelLV:
I have the same problem. I thought my card was broken but I see that it happens to more people.

humpedli:
I just ordered 8 piece of Sandisk SD (16 GB) card for my 8 cameras. Sometimes I experience the same issue, but only on the same camera and not always, other SD cards / cameras are working well. Anyway it's an interesting issue.

ravensteijn:
I got the same issue. It helped me connecting the sd card back to my computer. i then got a message that the disk needs repair. When i hit repair disk it was solved. maybe this helps?

pascal1n:
same issue here ! 
 a special linux configuration is nedded ?? (like no log on SD or something like that ?
Thanks for help

Gamer106:
Same issue for me. I use a smart-switch to turn the camera off when I’m home and on when I leave. I can’t remove the SD card every time this happens due to the install locations, so it’s a pain in the neck to resolve by having windows scan/fix the disk.

`[   20.999410] FAT-fs (mmcblk0p1): error, clusters badly computed (12 != 11)`
`[   21.006556] FAT-fs (mmcblk0p1): Filesystem has been set read-only`

I guess it’s writing to the disk constantly and pulling power is a no-no.

jmtatsch:
Why not just toggling motion detection via mqtt?

Gamer106:
I may just leave them powered on now that they’re local only. I was told the cameras on factory firmware sometimes remain “on” while you tell them to turn off so I used a switch.

jmtatsch:
That is very true. I don't trust (especially chinese) cloud services at all.

jvence:
[...]

Issue #731, opened 2018-10-09
ext3/4 support


arcaine2:
Does Dagang support ext3/4 as external media? I have a FAT32 formatted pendrive (brand-new) connected to USB port and enabled timelapse to save JPGs there. This works fine but there are always some issues while deleting old images (tons of them each day, deleting them takes a longer while) that lead to filesystem corruption. The more data to delete, the more often it corrupts, switch filesystem to read-only mode what have to be fixed manually. fsck on the camera often segfaults (possibly not enough free RAM). I recently switched from pictures based timelapse to my own recording script that records one hour long video and switches to a new file every hour. This gives me 24 files per day but there are still times when filesystem switches to ro mode while removing old files. Maybe ext filesystem would be more stable but i don't see required mkfs utility available.

arcaine2:
Sure, but don't judge, it needs some rewriting. I essentially modified some other script:

```
#!/bin/sh                                                                                                                                                                                                  
PIDFILE="/run/recording.pid"                                                                                                                                                                               
BASE_SAVE_DIR="/system/media/pendrive/DCIM/VIDEO"                                                                                                                                                          
                                                                                                                                                                                                           
SAVE_DIR_PER_DAY=1                                                                                                                                                                                         
SAVE_DIR_PER_HOUR=0                                                                                                                                                                                        
AUTO_DELETE_OLD_RECORDINGS_DAYS=5                                                                                                                                                                          
SLEEP_INTERVAL=1                                                                                                                                                                                           
                                                                                                                                                                                                           
                                                                                                                                                                                                           
if [ ! -d "$BASE_SAVE_DIR" ]; then                                                                                                                                                                         
  mkdir -p $BASE_SAVE_DIR                                                                                                                                                                                  
fi                                                                                                                                                                                                         
                                                                                                                                                                                                           
status()                                                                                                                                                                                                   
{                                                                                                                                                                                                          
  pid="$(cat "$PIDFILE" 2>/dev/null)"                                                                                                                                                                      
  if [ "$pid" ]; then                                                                                                                                                                                      
    kill -0 "$pid" >/dev/null && echo "PID: $pid" || return 1                                                                                                                                              
  fi                                                                                                                                                                                                       
}                                                                                                                                                                                                          
                                                                                                                                                                                                           
start()                                                                                                                                                                                                    
{                                                                                                                                                                                                          
  if [ -f $PIDFILE ]; then                                                                                                                                                                                 
[...]

Issue #723, opened 2018-10-04
Error starting update process


jdbeitz:
Running Version: 2018-09-05T06:32:31Z

When i log into the camera (WyzeV2) and try to update the software, i receive:
Error starting update process

Question - am i supposed to be able to do an in place update here?

ortegafernando:
hi @jdbeitz , did you find a solution ?

ortegafernando:
Finally, mine was a problem realated with read only sdcard

Issue #648, opened 2018-08-21
general feedback - delay and motor, etc.


TungstenE2:
Hi all,

first of all I was very impressed about the hack of the Dafang cam.
In general about the option and the active project on this.
But after several days of testing and playing around I am a bit disappointed,
but I am not sure if it is about the hardware or the software of the hack.
Tested with GUI, VLC and Tinycam Pro.

Here comes a list of my findings:

- live video has a strong delay and only about 5fps instead of 25fps as defined in settings
- error message on read-only sd-card
- auto-night modus does not turn on ir-leds
- motor does not turn correct, returns back instead of following orders and is stuttering
- several settings are not implemented in GUI. while this is a lot of effort for sure, a note in the 
  GUI would be helpfull to check the config files for all settings.
- only one-way audio from camera is supported

So my main issues are the delay/5fps and the not correct working motor.
How to fix this or am I doing something wrong?

thx

eridesktop:
Playing with Video Size, Video Format and Bit Rate in Video Settings Option maybe can help you solve the FPS problem
My settings was Video Size 1600 x 900 , Video Format SMART and Bit Rate 2000, of course you can try another option, just trial and error and find which option suites you the best.

tsvi:
To all those with this problem.
Could you please check if you stop the RTSP server and then run:

```
dosfsck -a /dev/mmcblk0p1
mount -o remount,rw /system/sdcard
/system/sdcard/autoupdate.sh
```
If the problem stops occuring?

TungstenE2:
good to see someone is working on the issues. but I can not test this as I sold my dafang, as it was not reliable working. but I will watch  the progress and may be buy one again if issues are fixed. thx

PR #899, opened 2019-01-29
 Log by default to /tmp so as to avoid SD card corruption

tsvi:
I discovered that the errors I was getting about the SD card similar to #880 were always related to the v4l2rtspserver log. It seems to me that especially when turning detection tracking due to the high number of writes, writing to the log causes SD card corruption. I suppose that the SD card controller is not built for such a speed. Since changing logging to log to /tmp which is mounted as tmpfs, I hadn't had any issues.

Should fix most cases of read-only partition, see: #880, #409, #648, #723, #731, #396, #457 after running `dosfsck -a /dev/mmcblk0p1`

### DIFF
--- a/firmware_mod/controlscripts/rtsp-h264
+++ b/firmware_mod/controlscripts/rtsp-h264
@@ -1,6 +1,6 @@
 #!/bin/sh
 PIDFILE="/run/v4l2rtspserver-master-h264.pid"
-LOGDIR="/tmp/log"
+LOGDIR="/tmp"
 LOGPATH="$LOGDIR/v4l2rtspserver-master.log"
 export LD_LIBRARY_PATH='/system/sdcard/lib/:/thirdlib:/system/lib'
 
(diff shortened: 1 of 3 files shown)


mikeage:
What happens when the tmpfs runs out of space, which will happen far before the SD card will? IIRC, there's no log rotation here.

jmtatsch:
Thanks.

tsvi:
@mikeage so we should implement a homemade version of log rotation. Don't have the time for this but according to some research I did today, I think we should do a weekly or daily `mv /tmp/v4l2rtspserver-master.log /system/sdcard/log/v4l2rtspserver-master.log` and then send the SIGUSR1 signal to the v4l2rtspserver, so it resets the pointer to 0 in the log.

But again this can help all the cases of read-only sdcard (hopefully).

jmtatsch:
A homemade version of log rotate would be appreciated as well of course.

PR #901, opened 2019-01-30
Improved time zone selection

### LINKED ISSUE #95: Date/time sync
As in the last version i mentioned the time sync works, but ...

For me (german) it should be 'CET+1' (summertime additional 1 hour at the moment), but when entering CET+1 i'm two hours behind. When entering 'CET-1' i get the correct time for my timezone.

### LINKED ISSUE #423: Time Zone Issue
Setting the time zone to GMT-2 corrects the time being displayed, though strangely I am in GMT+1 (Netherlands) so it anything it could either be GMT+1 or GMT+2 (daylight saving), but for sure not GMT-2:

image

For sanity check I tried different NTP severs (0.europe.pool.ntp.org).

### LINKED ISSUE #586: timezone settings
Hi!

It looks to me that there is something wrong with the timezone setting.

First of all it seems to be adding / subtracting in the wrong direction.

Like here for Norway it should be currently UTC+2 but to get the correct time shown UTC-2:00 needs to be entered.

timezone setting screen shot 2018-07-20 at 12 33 17

further
* auto daylight savings should be present 
* support for other named time zones, like CET or Europe/Oslo should be available

### LINKED ISSUE #630: Correct timezone (TZ) settings -- submit your settings here
Hopefully this is the thread to end all timezone setting threads: Here is what I know so far about correct TZ values for the webUI TZ setting (which btw.  reads from and and saves to the file  /system/sdcard/etc/TZ ):

#US eastern standard time, with daylight savings changeover programmed
EST+5EDT,M3.2.0/2,M11.1.0/2
#US pacific standard time, with daylight savings changeover programmed
PST+8PDT,M3.2.0/2,M11.1.0/2
#Central European Time, with summer time changeover programmed
CET-1CEST,M3.5.0/2,M10.5.0/3

Why these complicated expressions, you may ask?  Well, part of the complexity is the programming of daylight saving time changeover dates. Also, people may find the signs of the numerical offsets counterintuitive. For example, EST+5 (plus) when EST is more typically thought of as GMT-5 or UTC-5 (minus). Likewise, CET is often thought of as UTC+1. Why the opposite sign in the TZ spec? Well, this is because the POSIX definition of TZ uses the opposite definition of offset than what most humans do: The POSIX offset is to be *subtracted* from GMT or UTC to get the local time.

Obviously, I did not invent this particular wheel. Other people have hinted/stated various bits and pieces of this, but it took a long time for me to understand what was going on. Hence I wanted to contribute a be-all-and-end-all thread where the correct settings can be explained and collected up front.

Ideally, the webui would at least contain some quick examples/information about the offset definition in POSIX being the opposite sign of the "human" definition, and a link to the POSIX TZ reference material, as included below here. Maybe even a link to this thread, in the webui, would be useful as well!!??

Reference:
https://www.gnu.org/software/libc/manual/html_node/TZ-Variable.html

### LINKED ISSUE #683: Timezone calculation inverted
UTC gives me 20:00:00
if I change my timezone to UTC-3 it gives me 23:00:00
but if I change my timezone to UTC+3 it gives me 17:00:00

### LINKED ISSUE #814: OSD Timezone
Currently looks like the OSD only shows the time in UTC. Can not find a way to change it to the local TZ:

tz config
tz osd

JelmerT:
Setting the time zone was pretty confusing (see #586, #95, #423, #683, #814, #630).

I replaced the text field with a drop down list and filled it with time zone entries based on the [openwrt list](https://openwrt.org/docs/guide-user/base-system/system_configuration#time_zones) (not sure how this part is licensed).

You can now just select your time zone out of the list instead of doing weird calculations.

Not sure if `timezones.tsv` is in a good place. There's also a new config file `timezone.conf` generated after you selected a timezone for the first time. 
`/etc/TZ` is still where the system gets its time zone setting.

fixes #586, fixes #95, fixes #423, fixes #683, fixes #814, fixes #630 and probably others


JelmerT:
This might need a default `timezone.conf` file, the drop down list might show the wrong name on a clean install. I'm not sure what the default TZ is set at, and don't feel like wiping my cam.

jmtatsch:
I agree that we should have a default timezone.conf. I would vote for good old GMT0 as default timezone.

jmtatsch:
Can you add a timezone.conf please, so this can be merged?

JelmerT:
The default is as of yet untested on a clean install. If I get some time today I'll reflash my cam.

JelmerT:
Might be worth documenting somewhere that `/config/timezone.conf` just changes the timezone displayed in the web interface and not the actual system time. That's still in `/etc/TZ`. 
Maybe that config file needs a better name?

jmtatsch:
I would say fair enough :)
Thank you for fixing that long standing issue.

JelmerT:
So the problem with picking `Europe/London` as default is that it doesn't match the default `GMT0`.
As the timezone of London is apparently `GMT0BST,M3.5.0/1,M10.5.0`, something to do with summer and winter time I assume.
So now on a clean install the displayed timezone won't match the system timezone until you submit that form.

PR #903, opened 2019-01-30
Added info panel under manage -> settings

JelmerT:
It started with just uptime, because who doesn't want to see how long their internet connected thing has been alive? 
But it ended with a new info panel.

Not sure where I can get the model name though, any ideas?
It feels like the whole menu system might benefit from an overhaul, to place things in a more logical fashion. Info with info, settings with settings, services ...


jmtatsch:
You can definitely give the menu ordering an overhaul ...

PR #908, opened 2019-02-01
Basic model detection

JelmerT:
Added a model detection script since `hostname -i` didn't do much.

I only have a Dafang, so needs to be tested on different models.
Is there a way to detect the Wyzecam Pan?


JelmerT:
done

PR #910, opened 2019-02-02
web interface menu reshuffel

JelmerT:
First go at a clean-up of the web interface. Heavily influenced by the OpenWRT menu structure.

Needs a little more cleanup, but this is a good start. Some menus will be expanded with more functions (like turn on/off ftp, change ssh pass, ...)
Will probably move network to logs, since there is no config for network available.
Also removed missile launcher stuff, since that hardware seems to be really old and unavailable.

I can rebase once the other PRs are pulled.


jmtatsch:
Looks much more structured already, thank you so much! Can you remove the bit about starting the original firmware, it’s not working anymore. And then rebase....

JelmerT:
I think this should do it

PR #911, opened 2019-02-02
MQTT message and control fix for RTSP flipped case

sendorm:
The mqqt controls were reversed for the rtsp flipped case. This PR fixes that.


sendorm:
Yes I just left the old code for non flipped case and tested for flipped and non-flipped cases both.

PR #912, opened 2019-02-02
Add option for static IP configuration

harisokanovic:
Configure static IP and netmask when config/staticip.conf is present.
Add config/staticip.conf.dist example.

Additionally, configure a static resolv.conf when config/resolv.conf
is also present (in addition to config/staticip.conf). This is normally
configured by DHCP. Add config/resolv.conf.dist example.

Tested both DHCP and static IP configs over Wifi on Wyze Cam V2.

### DIFF
--- a/firmware_mod/config/resolv.conf.dist
+++ b/firmware_mod/config/resolv.conf.dist
@@ -1,0 +1,20 @@
+## An example resolv.conf to use with static IP configuration
+## This file is only relevant when a static address is configured
+## Otherwise, resolv.conf is managed by DHCP
+## Uncomment lines as needed, everything is optional
+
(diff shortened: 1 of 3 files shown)


JelmerT:
Nice work, consider expanding the overhauled web interface too (#910) , so you can configure this via more user friendly settings under the new `network` menu.

jmtatsch:
Great addition. Thank you for helping out.

harisokanovic:
@jmtatsch: Thanks for the pull. This alt firmware for my Wyze cams has saved me considerable time. The least I can do is upstream features I'm adding!

@JelmerT: I'm not that familiar with the web interface. I'll give it a shot if I start using it.

PR #913, opened 2019-02-02
Cleanup detectionOn.sh to improve performance

harisokanovic:
Modify as follows:
 1. Run `getimage` once at beginning of script, save image to /tmp
 2. Launch all notification tasks (save snapshot, mqtt, email, telegram,
    and user scripts) as background jobs, use same image image in all
    cases for consistency
 3. Wait for all background jobs to finish before deleting tempfile

In motion.conf.dist:
 * Add default debug_msg_enable=false
 * Fix typo: s/sendemail/send_email

Testing: Ran detectionOn.sh with all options+debugging enabled on a
Wyze Cam V2 and verified the following:
 * LED triggered
 * Snapshot jpg saved to sdcard
 * Old snapshots removed from sdcard
 * Ran few user scripts to read tempfile
 * sendPictureMail.sh, mosquitto_pub.bin, and telegram ran, but I
   didn't verify delivery

### DIFF
--- a/firmware_mod/config/motion.conf.dist
+++ b/firmware_mod/config/motion.conf.dist
@@ -2,6 +2,9 @@
 # edit this file and move it to /system/sdcard/config/motion.conf #
 ############################################################
 
+# Enable verbose debug messages to stderr 
+debug_msg_enable=false
+
 # Configure the motion core parameters:
 
 region_of_interest=0,0,1280,720
@@ -20,7 +23,7 @@ publish_mqtt_message=false
 publish_mqtt_snapshot=false
 # Save the image locally
 save_snapshot=false
-sendemail=false
+send_email=false
 send_telegram=false
 save_dir=/system/sdcard/motion/stills
 save_file_date_pattern="+%d-%m-%Y_%H.%M.%S"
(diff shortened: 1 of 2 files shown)


jmtatsch:
Thank you for these improvements.

saipsa:
@jmtatsch This PR blocks the script, since it waits for all jobs to be completed before exiting the script. The wait should be done in background.

saipsa:
@jmtatsch If there is network delay, snapshots will not be taken on motiondetection due to blocking.

harisokanovic:
@saipsa that's a good point, but the solution can't be blind forking because that can eat up memory. Last I checked, my Wyze V2 has something like 30M of working ram after the kernel, getty, ntpd, ftpd, sshd, rtspd, lighttpd, etc are loaded. There's not a lot to spare!

We might consider per-service blocking. So for example, if sendmail stalls, we'll avoid it on the next motion event while other services get to run. This can be implemented with flocks; see PR#922 for an example.

saipsa:
@harisokanovic I think if we store snaps when there is delay or loss of connection and try sending them on regaining connection or as the queue clears that would be ideal.

harisokanovic:
@saipsa  That's not a bad idea. For example, you could queue up images on the sdcard when `save_snapshot=true` and run one instance of each notifier as fast as it can go. It's a bit trickier to implement though.

saipsa:
@harisokanovic we can store queued filenames in text file and remove filenames as they are processed.

PR #914, opened 2019-02-02
Cleaned up update page

JelmerT:
Update page looks a little cleaner. I removed the github login references, it's kind of weird/bad idea to fill out your github password on a probably not so secure IoT device. And it seems to work just fine without.


JelmerT:
Maybe a confirm alert box could be added, not sure.

PR #920, opened 2019-02-03
add default gateway

i4s-pserrano:
Configure default gateway in case of define static ip

### DIFF
--- a/firmware_mod/run.sh
+++ b/firmware_mod/run.sh
@@ -141,6 +141,12 @@ if [ -f "$CONFIGPATH/staticip.conf" ]; then
   staticip_and_netmask=$(cat "$CONFIGPATH/staticip.conf" | grep -v "^$" | grep -v "^#")
   ifconfig "$network_interface_name" $staticip_and_netmask
   ifconifg "$network_interface_name" up
+  # Configure default gateway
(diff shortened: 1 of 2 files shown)


firmware_mod/run.sh
```diff
@@ -141,6 +141,12 @@ if [ -f "$CONFIGPATH/staticip.conf" ]; then
   staticip_and_netmask=$(cat "$CONFIGPATH/staticip.conf" | grep -v "^$" | grep -v "^#")
   ifconfig "$network_interface_name" $staticip_and_netmask
   ifconifg "$network_interface_name" up
+  # Configure default gateway
+  if [ -f "$CONFIGPATH/defaultgw.conf" ]; then
+    defaultgw=$(cat "$CONFIGPATH/defaultgw.conf" | grep -v "^$" | grep -v "^#")
+    route add default gw $defaultgw wlan0
```
harisokanovic: You probably should use `"$network_interface_name"` instead of `wlan0`. I think there's some camera variant with a physical Ethernet connection, which would be named eth0 instead of wlan0. That variable is set appropriately a few lines above.

i4s-pserrano: Ok, I make this change

PR #922, opened 2019-02-04
detectionOn.sh: Upload snapshot and/or video stream to ftp server

harisokanovic:
Enabling `ftp_snapshot=true` in motion.conf uploads a single image to
ftp server on each motion event, similar to other handlers.

Enabling `ftp_video=true` starts a background task to stream 1fps video
until motion stops. It minimally steams for 10 seconds (configurable),
refreshing the countdown as motion continues.

### DIFF
--- a/firmware_mod/config/motion.conf.dist
+++ b/firmware_mod/config/motion.conf.dist
@@ -27,3 +27,14 @@ send_email=false
 send_telegram=false
 save_dir=/system/sdcard/motion/stills
 save_file_date_pattern="+%d-%m-%Y_%H.%M.%S"
+
+# Configure FTP snapshots and videos
+ftp_snapshot=false
+ftp_video=false
+ftp_video_duration=10
+ftp_host="ftp.myserver.net"
+ftp_port=21
+ftp_username="user1"
+ftp_password="password2"
+ftp_stills_dir="motion/stills"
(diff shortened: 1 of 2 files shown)


jmtatsch:
Love it.

PasteurPT:
Almost perfect! Thanks a lot! 

Can I ask you to make some changes? Number of snapshots to upload to FTP and more FPS in video stream? I'm trying to do that but without success.

harisokanovic:
I also had trouble reliably getting higher frame rate, hence the shoddy jpeg stitching approach. I would love to simply replace it with rtsp, but that was incredibly unreliable in my experiments. It took around 10 seconds to negotiate a connection with local rtsp server.

I'll be sure post my changes if figure it out.

harisokanovic:
I don't think the camera has a full-featured FTP client, it can only "put" and "get" files. You should configure your FTP server to clean up old files instead, e.g. schedule a cron job.

jmtatsch:
@harisokanovic do you think you could make the video recording more generic so that video can also be sent via email or telegram?

pootmhicnought:
Also, might want to add this to motion.conf.dist:

# seconds of video to record when ftp_video=true.
ftp_video_duration=10

Without that, I was getting 0 byte videos.  Seems to not have been defaulted anywhere.

PR #924, opened 2019-02-05
Use a more standard string comparison

nbarrientos:
This also silently implements correctly the default behaviour when SENDLDR is not set which is to send the value. Sorry *wink*


nbarrientos:
@jmtatsch

firmware_mod/scripts/mqtt-autodiscovery.sh
```diff
@@ -36,7 +36,7 @@ VER="Dafang Hacks"
 /system/sdcard/bin/mosquitto_pub.bin -h "$HOST" -p "$PORT" -u "$USER" -P "$PASS" -t "$AUTODISCOVERY_PREFIX/switch/$DEVICE_NAME/ir_cut/config" ${MOSQUITTOPUBOPTS} ${MOSQUITTOOPTS} -r -m "{\"name\": \"$DEVICE_NAME ir filter\", \"unique_id\": \"$MAC_SIMPLE-ir-filter\", \"device\": {\"identifiers\": \"$MAC_SIMPLE\", \"connections\": [[\"mac\", \"$MAC\"]], \"manufacturer\": \"$MANUFACTURER\", \"model\": \"$MODEL\", \"name\": \"$MANUFACTURER $MODEL\", \"sw_version\": \"$VER\"}, \"icon\": \"mdi:image-filter-black-white\", \"state_topic\": \"$TOPIC/ir_cut\", \"command_topic\": \"$TOPIC/ir_cut/set\"}"
 
 # Light Sensor
-if [ ! $SENDLDR == "false" ]; then
+if [ "$SENDLDR" != "false" ]; then
```
jmtatsch: Why double negation? 
isn't that even easier to understand ?
if [ "$SENDLDR" == "true" ]

nbarrientos: To avoid a behaviour change for users that update the software carrying forward an existing `mqtt.conf` where `$SENDLDR` is not set. If we did what you're proposing lots of people would start seeing that the LDR status is not being sent after updating.

Alternatively we could do:

```bash
if [ "$SENDLDR" = "" -o "$SENDLDR" = "true" ]; then
```

Which looks even uglier IMO but I'm happy to change it :)

PR #926, opened 2019-02-06
Optimize tracking detection

tsvi:
Put motor movement in background
Skip sleep after PTZ script. Sleep is already taken care of in the script.
Fix case of 2 regions detected.
Add sleep after movement


jmtatsch:
thanks

PR #927, opened 2019-02-06
Pan tilt calibrate on boot

JelmerT:
Make sure we got the correct position value from the get-go. Original firmware also calibrates on boot.

### DIFF
--- a/firmware_mod/run.sh
+++ b/firmware_mod/run.sh
@@ -187,6 +187,9 @@ blue_led on
 ## Load motor driver module:
 insmod /driver/sample_motor.ko
 
+## Calibrate pan/tilt
+motor reset_pos_count
+
 ## Determine the image sensor model:
 insmod /system/sdcard/driver/sinfo.ko
 echo 1 >/proc/jz/sinfo/info


saipsa:
This should be off by default. I have camera mounted in such a way it can't rotate.

jmtatsch:
Also calibration doesn’t really work for cameras without endstops. It will move left right in an unbalanced way and windup the cables inside.

saipsa:
Then in that case, the best course of action would be to have startup-scripts folder and execute scripts in that folder. Folks who are interested in calibration, can put the script in that folder.

jmtatsch:
Or we could make calibration service and autostart it ...

JelmerT:
@jmtatsch there are commercially available cameras without end-stops? That seems like not a great idea 😄 

I'll try to pour it into a service and put it as autorun by default. People with HW restrictions/scheduled reboots could then disable the autorun part.
I don't see a solution for hardware without end-stops, except for detecting that specific hardware.

I'm pushing on the autorun by default since the calibrate button or service is pretty far away in some menu and for first time users it might take a while to find. So there is a chance they'll open an issue here first before they go looking for the calibrate service, which I want to prevent.

saipsa:
@JelmerT 

>  People with HW restrictions/scheduled reboots could then disable the autorun part.

Why not disable it by default and interested users can enable it?

saipsa:
>  they'll open an issue here first before they go looking for the calibrate service, which I want to prevent.

Without end stop, the wiring may come off. That would be catastrophic failure.

JelmerT:
Only now discovering previous problems with endstops (#287), have to read up on it and come back to this.

JelmerT:
So if I understand correctly: Earlier versions of Dafang HW had 'soft end-stops' (aka switches) and later HW have 'hard end-stops' (aka mechanical end-stop).
Earlier HW was calibrated by waiting for the end-stop switch to short and newer hardware is calibrated 'the cheap way' by running the motor into the end-stop for a full revolution.
Now if you try running an earlier HW model into a non-existing hard end-stop, you get trouble since you might miss the soft end-stop and there is no physical stop.

But I thought these problems were fixed with 64228e7b94a8bc99d1fa3025036bda49fb8bfbe9 and 260720dc84b4da5496d95f9eb27d9713cad51ad5. I can't find the source of the new `motor` binary,  but it should be possible to support both modes at the same time. (which it does right now afaik?)

jmtatsch:
There is no physical endstop to run into. It can run freely until the cable unplugs from winding up. Source should be in dafang main.

saipsa:
I have created PR for startup userscripts. You can place this script in startup folder. This is better than creating service that is disabled by default.

If user of camera without physical endstop (most of them won't know that in advance or the fact that motion calibration may destroy their camera), switches on the calibration service(out of curiosity or otherwise), it will destroy their camera.

JelmerT:
Added new pictures of the un-populated end-stop switches on new dafang HW in #932. This was already known, but just to confirm and document.

JelmerT:
> If user of camera without physical endstop (most of them won't know that their cam doesn't have physical endstop or the fact that motion calibration may destroy their camera), switches on the calibration service(out of curiosity or otherwise), it will destroy their camera.

You're telling me that people with the old HW can push the current "Calibrate" button in the web-interface and destroy their camera (aka have the motor pull out the connector)?
Shouldn't we fix that problem instead of disabling movement all-together?

JelmerT:
Are the unpopulated switches (J10) Normally Closed?

If yes, we could detect if they are present and based on that load the correct motor module.
[This](https://github.com/Dafang-Hacks/drivers/blob/16770a7c7f67520b85f8120adf9e0035357e7c91/misc/sample_motor/motor.c) one for new model without switches and [this](https://github.com/Dafang-Hacks/drivers/blob/16770a7c7f67520b85f8120adf9e0035357e7c91/misc/sample_motor2/motor.c) one for old model with switches.

jmtatsch:
> You're telling me that people with the old HW can push the current "Calibrate" button in the web-interface and destroy their camera (aka have the motor pull out the connector)?
People with the new HW can potentially destroy their devices. Would probably need them calibrating 100x or so.

We should definitely fix this. I don't know if they are NC or NO. Only know this:
https://github.com/EliasKotlyar/Xiaomi-Dafang-Hacks/issues/287#issuecomment-384582924
I think it gpis goes 1 on old hw once the endstaop is reached.

saipsa:
> You're telling me that people with the old HW can push the current "Calibrate" button in the web-interface and destroy their camera 

At present calibration will happen only once, in case of service, it will happen every time the cam boots. Then the probability of damage  -> 1, as it will continue calibrating until mechanical failure occurs. 

> (aka have the motor pull out the connector)?

That is the best case scenario of failure. Short-circuit is another.

MasterPIC:
In the meantime I reverted to an old motor bin and driver set to have calibration working perfectly on my older Dafang with endstops switches. I disabled X and Y OSD report: there is no position report...

MasterPIC:
I just created a pull request (#937) to share the solution I'm preferring for my old Dafang equipped with endstops.
I think a good idea could be allowing the user to select the preferred solution according to the hardware in use.

PR #929, opened 2019-02-07
Fix mqtt motion sensor autodiscovery

liads:
Seems like the "icon" property was never supported for the mqtt binary_sensor and version 0.87 of Home Assistant now fails to add it.


nickrout:
Confirm the error, would be nice to get this fixed :)

jmtatsch:
It was working for me before but they must have changed something.

mikeage:
Nice work. This was a breaking change in 0.87; it worked with 0.86.4.

Note that there's another issue with parsing the image sent over MQTT: https://github.com/home-assistant/home-assistant/issues/20855 [sorry for hijacking this thread; this is mostly just an FYI I guess]

liads:
According to the release notes, version 0.87 changed how unknown settings for mqtt entities are handled. They were ignored before and now they're treated as errors. I think that's why it used to work.

jmtatsch:
So should we work to make icon a known setting?

liads:
I don't think it's really needed. The binary_sensor has a device_class property to indicate which icons (and status text) should be used. There's also a different icon depending on the state, so I guess that's why this property was excluded.

vivekmreddy:
I'm having an issue that may be related to this. Is anyone having a problem where if you toggle the motion detection with mqtt it defaults to the "very high" sensitivity?

PR #937, opened 2019-02-11
Please don't merge - Support for Dafang with endstops

MasterPIC:
Hi,
I reverted to an older motor module by @EliasKotlyar ( https://github.com/EliasKotlyar/Xiaomi-Dafang-Hacks/commit/b30164e178e7c4d2bd740f0e08540fc4a758f891 )
Calibration works as expected by using the (required) endstops.
I edited common_functions.sh to get position too...
I think a good idea could be allowing the user to select the preferred solution according to the hardware in use.


Kaan88:
Wow, now it calibrates so smoothly. Funny things is it seems official firmware doesn't take advantage of the sensors either.

Kaan88:
If you are still using your Dafang, I just discovered that sample_motor.ko from firmware 5.5.1.177 has sensors working. No need to replace any scripts, or motor.bin. It just works as it should.

MasterPIC:
> If you are still using your Dafang, I just discovered that sample_motor.ko from firmware 5.5.1.177 has sensors working. No need to replace any scripts, or motor.bin. It just works as it should.

My mod makes Dafang reverting to the last position after a powercycle.

Kaan88:
I obviously know that, important thing about your mod was that it used the sensors, but your mod is not compatible with latest release. But with just a driver change you can have everything, including reverting to last position since official already has that.

MasterPIC:
I just followed your indications and I see it works as expected with the use of the endstops and no need to recalibrate after a powercycle.
Thank You!

Issue #940, opened 2019-02-14
Motion Detection -  MQTT Motion Detection "ON" calls function with static motion sensitivity of "4" instead of what is in the motion config or last value


vivekmreddy:
Hi all,

This is my first post on github (long time lurker), and I noticed an issue that I've dealt with in a poor manner on my own setup (just changed the value from 4 to 1) but I'm not confident enough to add a change to the code where it dynamically reads the value from the configuration.

Long story short, I use MQTT with Home Assistant to enable/disable motion detection on my camera. After enabling again, it always reverted back to the highest sensitivity which threw all sorts of false positives. It looks like that mqtt-control.sh is calling the motion_detection function in common_functions.sh. The line in the function, that pertains to enabling/disabling motion:

`/system/sdcard/bin/setconf -k m -v 4`

located in: Xiaomi-Dafang-Hacks/firmware_mod/scripts/common_functions.sh at line 305

I'm pretty confident it's possible to pull the value from the the motion.conf, but since I'm a bit new, and I didn't see anything out there on this issue so I figured I'd at least mention it so someone more capable than I, who could make the change quicker.

Please let me know if you have any more questions.

jmtatsch:
Should work now. Please report back if not fixed.

vivekmreddy:
Thanks for updating! But it seems the enable component doesn't work. I can disable the motion detection, but MQTT cannot re-enable it.

vivekmreddy:
Just to confirm, I removed that line from my file, and it works as expected. Thank you!

I will read up on how to submit fixes for the next time. If there is any good documentation, besides the official, that is worth reading, I'm all ears. Thanks again!

PR #941, opened 2019-02-14
Stop re-write of motion.conf

vivekmreddy:
Removed line that rewrites "-1" to motion_sensitivity in motion.conf, when using motion_detection "off". Since the motion_detection function pulls motion_sensitivity from the motion.conf file when running motion_detection "on", it will never re-enable motion detection after the first time you disable it since "-1" is written to the motion.conf. #940

### DIFF
--- a/firmware_mod/scripts/common_functions.sh
+++ b/firmware_mod/scripts/common_functions.sh
@@ -314,8 +314,6 @@ motion_detection(){
     ;;
   off)
     /system/sdcard/bin/setconf -k m -v -1
-    rewrite_config /system/sdcard/config/motion.conf motion_sensitivity "-1"
-
     ;;
   status)
     status=$(/system/sdcard/bin/setconf -g m 2>/dev/null)


jmtatsch:
Thank you for learning how to create a pull request.
I am unsure however about the proposed change.
If motion is being disabled, that change should also be persisted to the motion.conf file.

vivekmreddy:
Does it make sense to add a line item “default_motion_sensitivity” in the conf and use that as the read value for the motion_detection “on/off”? Then you can still write the -1 state to the conf.

-Vivek R.

tboyce021:
I just ran into this as well. I think it would make sense to have separate "sensitivity" and "on/off" settings, e.g.
```
motion_sensitivity=4
motion_detection=on
```
That lets you easily enable/disable the detection while always retaining the desired sensitivity when enabled. It's pretty similar to default_motion_sensitivity, but makes things a little clearer, in my opinion. I haven't looked into the scripts to see if that would be a lot more work though.

jmtatsch:
Ah okay, now I understand why this is needed. I would prefer the cleaner first option.

vivekmreddy:
> I just ran into this as well. I think it would make sense to have separate "sensitivity" and "on/off" settings, e.g.
> 
> ```
> motion_sensitivity=4
> motion_detection=on
> ```
> That lets you easily enable/disable the detection while always retaining the desired sensitivity when enabled. It's pretty similar to default_motion_sensitivity, but makes things a little clearer, in my opinion. I haven't looked into the scripts to see if that would be a lot more work though.
> 
> Edit: I looked into it briefly and it does seem harder to do that. I tweaked this a bit to just save the current value to `default_sensitivity` before turning it off and then restore that and rewrite to `motion_sensitivity` again when turning on. Seems to work fine so far.

That makes sense to me.

jmtatsch:
Implemented your suggestions in 
https://github.com/EliasKotlyar/Xiaomi-Dafang-Hacks/commit/5d041e4f0522efbfdf3c80fedaaec633c6aca9c2

PR #951, opened 2019-02-22
Fixes sending strings using Matrix

xMateMCPE:
If you want to send a message using `matrix m message`, the sent message will only contain $text due to a problem with the code.


jmtatsch:
Thanks for fixing this.

PR #1039, opened 2019-05-05
Update estaticip.conf.dist to define netmask parameter

i4s-pserrano:
update staticip.conf.dist file to define correctly the netmask.


jmtatsch:
Thanks for contributing.

Issue #1042, opened 2019-05-07
Trigger MQTT autodiscovery on reconnection or manually


mikeage:
### Description

Currently, if the camera boots before the mqtt server is running (e.g., power failure), the autodiscovery message is lost (it fails to send, but that doesn't report any errors, and it doesn't retry).

There are two ways that I can think of to fix this:

1. Always send autodiscovery when we reconnect
2. Add a command to manually send the autodiscovery message; this can be a bit of a chicken and egg problem, but it could also list to (for example) `myhome/` rather than `myhome/foo`

I'll try to work on this, but before I do (I probably won't have any time for the next few days), any thoughts on which option is preferable?

jmtatsch:
1. Which reconnection do you mean? The MQTT client is frequently connecting/disconnecting every loop run. This is already a problem, and can break the video stream if to much MQTT control commands are coming in.

2. Better than nothing, but still annoying to do it manually :)

mikeage:
Good point about #1. Forgot about that. So it'd probably have to be option 2... I have my homeassistant automatically send out announce requests anyway, so it's not really manually. 

Another option would be to set a flag if the autodiscovery send fails, and to keep trying until it succeeds. I think we send it with `retain` set, right? If not, it probably should be, to avoid a problem if the camera comes up after MQTT but before HA does (or if HA is restarted).

mikeage:
Took me a while, but I just opened a PR. Works for me ;-)

PR #1051, opened 2019-05-15
Add MQTT command to resend autodiscovery

mikeage:
In some cases (e.g., power failure), it is possible for the camera to
boot before the MQTT server is running. In this case, the autodiscovery
message will be lost.

To address this, this commit adds a topic called $LOCATION/set which
accepts an argument of 'announce'. This will trigger resending of the
autodiscovery message. It does not use the standard path of
$LOCATION/$DEVICE_NAME, because it is intended for all Xiaomi-Dafang-Hacks
cameras.

To use this, once your MQTT server is up, publish a message to
$LOCATION/set with the payload 'announce'. In Home Assistant, this can be done
with the following automation:
```yaml
- id: '000'
    alias: Dafang Announce on startup
    trigger:
    - event: start
      platform: homeassistant
    action:
      service: mqtt.publish
      data:
        payload: announce
        topic: myhome/set
```

### DIFF
--- a/firmware_mod/scripts/mqtt-control.sh
+++ b/firmware_mod/scripts/mqtt-control.sh
@@ -6,8 +6,11 @@
 killall mosquitto_sub 2> /dev/null
 killall mosquitto_sub.bin 2> /dev/null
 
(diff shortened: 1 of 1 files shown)


mikeage:
Closes #1042

jmtatsch:
Thanks

PR #1086, opened 2019-06-28
Motion detectionOn script revamped; Added SMB

cadavre:
Hi, I've reviewed a little `detectionOn.sh` script and made some cleanups/addons.

This includes:

1. Added SMB upload
There is a native `smbclient` available so we can make use of it to upload files via SMB shares.

2. Revamped FTP upload for a simplified model
I believe "FTP video streaming" is a little overshot and it complicates the script a lot, making a duplicates on video recording. I've simply made it re-use `record_video` function.

3. Made some options generic
Filename pattern and video length had been made generic settings.

4. MQTT snapshot resize and optimization
MQTT max message size is 64KB, but `getimage` often creates bigger files. Especially when using 1080p mod. I added scaling to 1/2 and `jpegoptim` – for MQTT is should be enough for preview purposes.

5. Code cleanup
Some variables were reused not as they were named etc. Cleaned it up.

### DIFF
--- a/firmware_mod/config/motion.conf.dist
+++ b/firmware_mod/config/motion.conf.dist
@@ -24,24 +24,34 @@ send_telegram=false
 telegram_alert_type=image
 send_matrix=false
 
+# General
+file_date_pattern="+%d-%m-%Y_%H.%M.%S"
+video_duration=10
+
 # Snapshots
 save_snapshot=false
-save_dir=/media/motion/stills
+save_snapshot_dir=/media/motion/snapshots
 save_snapshot_attr="0666"
-save_file_date_pattern="+%d-%m-%Y_%H.%M.%S"
 max_snapshots=20
 
 # Videos
 save_video=false
 save_video_dir=/media/motion/videos
 save_video_attr="0666"
-video_duration=10
 max_videos=10
+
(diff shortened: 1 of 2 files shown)


harisokanovic:
I realize ftp streaming is somewhat complicated, but real-time off-camera recording has some useful security benefits. E.g. The camera will make partial recordings if interrupted during the process. A record-then-upload approach simply lose an entire clip when something goes wrong. Streaming also allows for much longer video clips that aren't bound by memory or flash availability, both of which are very limited on these cameras.

I prefer to keep this feature. It has no impact on the other simpler cases. We can move it into a different file if readability is an issue.

jmtatsch:
I agree with @harisokanovic. Security cameras are obviously the first thing to destroy if someone breaks in so a cloud backup is highly desired. So we shouldn't remove that feature.

jmtatsch:
I would however love to merge all your other proposals. Could you please update your pull request accordingly?

cadavre:
Oh, I missed somehow this PR. :)

@harisokanovic @jmtatsch – fair point for "ftp streaming".

I'll separate PRs, try to unify some things and introduce few new functionalities I've made last month.

eot

PR #1093, opened 2019-07-04
Support for groups and edited messages

m-pavel:
1. When bot is added to a group it receives commands like /command@botname instead of '/command' in direct chats.  Commit 4861496 allow bot to understand such commands.

2. When user edit message in telegram it comes to bot with unexpected json structure. This hangs up message processing at all. Commit de2bea6 add support for 'edited_message' processing.

### DIFF
--- a/firmware_mod/scripts/telegram-bot-daemon.sh
+++ b/firmware_mod/scripts/telegram-bot-daemon.sh
@@ -43,13 +43,13 @@ imageAlerts() {
 
 respond() {
   case $1 in
-    /mem) sendMem;;
-    /shot) sendShot;;
-    /on) detectionOn;;
-    /off) detectionOff;;
-    /textalerts) textAlerts;;
-    /imagealerts) imageAlerts;;
-    /help) $TELEGRAM m "######### Bot commands #########\n# /mem - show memory information\n# /shot - take a shot\n# /on - motion detect on\n# /off - motion detect off\n# /textalerts - Text alerts on motion detection\n# /imagealerts - Image alerts on motion detection";;
+    /mem*) sendMem;;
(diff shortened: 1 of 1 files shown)


firmware_mod/scripts/telegram-bot-daemon.sh
```diff
@@ -43,13 +43,13 @@ imageAlerts() {
 
 respond() {
```
znanev: What a coincidence! I was about to propose a similar pull request, because I implemented groups handling on my camera this morning. If _chatId_ is a negative number, this means it comes from a group (command arrives as **/command**@_botname_).

So, I used shell's native substring handling in order not to change individual case comparison patterns. I put following two statements in the beginning of the _respond()_ function, just before the case statement on line 45: 

```shell
cmd=$1
[ $chatId -lt 0 ] && cmd=${1%%@*}
```
Then the argument **$1** needs to be replaced with **$cmd** (lines 45 and 53).

As an added bonus, my proposed solution will correctly handle situations where commands coming from group chats appear to be substrings of another command (for example /mem will not be matched, if command sent was **/mem**bers).

jmtatsch:
Thank you for improving telegram support.

firmware_mod/scripts/telegram-bot-daemon.sh
jmtatsch: Can you make a pull request for your solution?

znanev: Sure!

I was out and about in the mountains with no internet access.. so back to reality now, here's the pull request:

[Pull Request #1](https://github.com/m-pavel/Xiaomi-Dafang-Hacks/pull/1)

PR #1104, opened 2019-07-18
Addet automatic Timelaps

Seg22:
You already got everything to capture the .jpg and convert it to .mov but you don't convert it automatically and i thought this is not what a timelaps should look like. So i changed timelaps and timelaps.sh

I´ve changed the way you name the .jpg so you don't have to convert it twice.
Then i´m converting the .jpg into .mov with date/time and deleting all .jpg so you can do more timelapses a day without doing something :)

### DIFF
--- a/firmware_mod/scripts/timelapse.sh
+++ b/firmware_mod/scripts/timelapse.sh
@@ -28,15 +28,15 @@ while true; do
     if [ ! -d "$SAVE_DIR" ]; then
         mkdir -p $SAVE_DIR
     fi
-    filename_prefix="$(date +%Y-%m-%d_%H%M%S)"
-    if [ "$filename_prefix" = "$last_prefix" ]; then
-        counter=$(($counter + 1))
-    else
-        counter=1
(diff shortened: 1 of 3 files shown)


firmware_mod/scripts/timelapse.sh
```diff
@@ -20,14 +20,15 @@ counter=0
...
+SAVE_DIR=$BASE_SAVE_DIR
     if [ $SAVE_DIR_PER_DAY -eq 1 ]; then
         SAVE_DIR="$BASE_SAVE_DIR/$(date +%Y-%m-%d)"
     fi
     if [ ! -d "$SAVE_DIR" ]; then
         mkdir -p $SAVE_DIR
     fi
+
+while true; do
     #filename_prefix="$(date +%Y-%m-%d_%H%M%S)"
```
bnilesh: Please increase the counter after while. With your changeset, I think image file will be overwritten.
    counter=$(($counter + 1))

Thanks!

PR #1106, opened 2019-07-22
Load `motion_timeout` setting from sdcard

ammaraskar:
Currently, restarting the camera will cause the `motion_timeout` setting to be defaulted to the last value. I think a call to set the config value in `configureMotion` was missed.


jmtatsch:
I think so too. Thanks for catching that.

PR #1133, opened 2019-08-26
 allow recording with credentials

chaoranxie:
if user uses username and password for the rtsp, then recording will fail. this fixes that


jmtatsch:
Thanks for fixing that.

PR #1135, opened 2019-08-27
Revert "Revert "Update mqtt-autodiscovery.sh""

zewelor:
Reverts EliasKotlyar/Xiaomi-Dafang-Hacks#1131

After verification in source code, looks like mqtt camera still doesn't support device attribute ( tested on newest HA 0.97.2 ). Docs ( https://www.home-assistant.io/components/camera.mqtt/#device ) states it should be supported but it gives error about unsupported key, and can't find any info about device in mqtt.camera code in HA. I think this should be reverted for now. I will try to add device in ha mqtt camera first and then it could be reenabled. Sorry for confusion.


jmtatsch:
Can you maybe open an issue with home assistant why its not working as it should?

zewelor:
I've already opened PR to add device to mqtt camera, lets hope it will be merged soon. https://github.com/home-assistant/home-assistant/pull/26238 .

PR #1155, opened 2019-09-07
Update flashinguboot.md

cf8geek:
added a disadvantage I didn't realize until it was too late

### DIFF
--- a/hacks/flashinguboot.md
+++ b/hacks/flashinguboot.md
@@ -12,6 +12,7 @@
 
 ## What are the disadvantages?
 If you flash the wrong u-boot, you can brick your device. I am not taking any responsibility for that!
+Also, you will be unable to revert to stock firmware by removing the microSD card.
 
 ## Requirements:
 


jmtatsch:
I don't think this is true. If no uEnv.bootfromsdcard.txt is found, it should default to boot from NAND also for stock fw.

cf8geek:
@jmtatsch I only know this because I inadvertently tested it. I could not get a demo.bin to load with the custom bootloader. If your testing yields different results, perhaps we have slightly different hardware?

jmtatsch:
@EliasKotlyar can you chime in on this?

jmtatsch:
Did you enable booting from NAND with an appropriate uEnv.txt in your sd root?
like mv uEnv.bootfromnand.txt uEnv.txt?

cf8geek:
No, but how does this help? If the ONLY file on the microSD card is demo.bin, it doesn't really matter if there was previously a .txt file there, does it? Also, which NAND does "bootfromnand" boot from?

jmtatsch:
Demo.bin is only flashed if there is nothing else on the sd card. Uboot boots from the nand address in uEnv.txt I think.

cf8geek:
My point being you CAN'T flash demo.bin from the microSD card after performing the bootloader replacement (hence this pull request). Or at least I couldn't on my two cameras. Which makes sense, if the bootloader is looking for uEnv.txt instead of demo.bin.

cf8geek:
Have you confirmed?

prawnz:
> @jmtatsch I only know this because I inadvertently tested it. I could not get a demo.bin to load with the custom bootloader. If your testing yields different results, perhaps we have slightly different hardware?

I am having this at the moment.. My Wyze Cam Pan can't even manually install any other firmwares now

cf8geek:
@prawnz Were you able to pull a working bootloader from another camera to remedy your problem? That's what I ended up doing.

prawnz:
@cf8geek nope I don't have another camera on the side to be able to do that.. Are you able to share your bootloader?

PR #1160, opened 2019-09-08
enable swap by config

i4s-pserrano:
configure swap by creating a swap config file. Lets you to update to new version and avoid the problem of revert changes at run.sh to swap=false


jmtatsch:
Great idea, thanks!

PR #1165, opened 2019-09-12
Adding isp_exposure value to mqtt-status?

exetico:
I'm not sure whether there's a better way, or not?

Also, i guess it could be an idea, to add it in mqtt-autodiscovery.sh, too.

But i'm not sure if it should be within the Light Sensor part, or another area? 

```
# Light Sensor
if [ "$SENDLDR" != "false" ]; then
  /system/sdcard/bin/mosquitto_pub.bin -h "$HOST" -p "$PORT" -u "$USER" -P "$PASS" -t "$AUTODISCOVERY_PREFIX/sensor/$DEVICE_NAME/ldr/config" ${MOSQUITTOPUBOPTS} ${MOSQUITTOOPTS} -r -m "{\"name\": \"$DEVICE_NAME light sensor\", \"unique_id\": \"$MAC_SIMPLE-light-sensor\", \"device\": {\"identifiers\": \"$MAC_SIMPLE\", \"connections\": [[\"mac\", \"$MAC\"]], \"manufacturer\": \"$MANUFACTURER\", \"model\": \"$MODEL\", \"name\": \"$MANUFACTURER $MODEL\", \"sw_version\": \"$VER\"}, \"icon\": \"mdi:brightness-5\", \"unit_of_measurement\": \"%\", \"state_topic\": \"$TOPIC/brightness\"}"
fi
```

### DIFF
--- a/firmware_mod/scripts/mqtt-status.sh
+++ b/firmware_mod/scripts/mqtt-status.sh
@@ -12,5 +12,6 @@ quality=$(/system/bin/iwconfig 2>/dev/null | grep "Quali" | sed -e "s/  *//")
 noise_level=$(echo "$quality" | awk '{ print $6}' | sed -e 's/.*=//' | sed -e 's/\/100/\%/')
 link_quality=$(echo "$quality" | awk '{ print $2}' | sed -e 's/.*=//' | sed -e 's/\/100/\%/')
 signal_level=$(echo "$quality" | awk '{ print $4}' | sed -e 's/.*=//' | sed -e 's/\/100/\%/')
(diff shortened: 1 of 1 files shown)


jmtatsch:
Looks good to me.
I will eventually change SENDLDR to parameter LIGHT_SENSOR with value either "hw" or "virtual"
For the autodiscovery maybe we should not call it ldr anymore but just more general light_sensor or exposure_sensor?

exetico:
HI again,

Take a look at:
https://github.com/EliasKotlyar/Xiaomi-Dafang-Hacks/pull/1166

Will that work out? I have not looked in the code before, so maybe i missed something..

Issue #998, opened 2019-04-03
TTF support not working with America/Los_Angeles


mauckc:
### Hardware

WCamPan v2

### Description

Setting up time zones on camera has no effect

### What I did to debug the issue

I updated the common_functions.sh script to manually enter in my specific POSIX date format.

```powershell
# Set timezone from the timezone config file to system timezone
set_timezone(){
  timezone_name=$(cat /system/sdcard/config/timezone.conf)
  timezone=$(/system/sdcard/bin/busybox awk -F '\t' -v tzn="$timezone_name" '($1==tzn) {print $2}' /system/sdcard/www/timezones.tsv)
  if [ "$(cat /etc/TZ)" != "$timezone" ]; then
    echo "$timezone" > /etc/TZ
  fi
```

to

```powershell
# Set timezone from the timezone config file to system timezone
set_timezone(){
  timezone_name=$(cat /system/sdcard/config/timezone.conf)
  timezone="PST8PDT" # HARDCODED IN THE POSIX TIMEZONE
  {print $2}' /system/sdcard/www/timezones.tsv)
  if [ "$(cat /etc/TZ)" != "$timezone" ]; then
    echo "$timezone" > /etc/TZ
  fi
```

### Evidence

I could see that the common_function.sh script was working correctly with the timezones.tsv file, but the actual system clock was not changing from UTC time.

This image shows the camera's "system" configuration when the clock cannot set to Pacific Standard Time.
![dafang_timezone_PST_error](https://user-images.githubusercontent.com/8161032/55502528-9833d300-5601-11e9-9f6b-06a6965615d2.PNG)

This image shows the camera's "system" configuration when I set clock to Pacific Standard Time manually.

![dafang_timezone_PST_success](https://user-images.githubusercontent.com/8161032/55502535-9d911d80-5601-11e9-81a1-a9f608170cc7.PNG)

Before hardcoding in my POSIX timezone, I attempted to update the my America/Los_Angeles POSIX date format in the timezones.tsv file and found that even though it would look correct on the settings page, it would not actually affect the camera's clock.  Maybe this issue is caused by the POSIX dates in the timezones.tsv file not being entered as a string?  Does this timezone setting work for other timezones?

impala454:
I tried to be a good citizen and figure this out and submit a pull request, but just ended up putting in a hack that worked for me.

I tried several combinations of things, ultimately this is what worked for me.  
[...]

PR #1170, opened 2019-09-18
Fixed chicago timezone

impala454:
It appears the customized POSIX formats do not work on this device (tested with `DAFANG`).  Changing the time zone to `America/Chicago` had no effect on the OSD.  When I removed the customized portion of the format it was resolved.  I'd suggest removing the rest of the customized formats but I wasn't planning to test them all so figured I'd just fix the one I know about.


impala454:
May also help resolve #998

PR #1177, opened 2019-09-29
Fixed the issue that auto starting recording could block the followin…

freegink:
…g services


jmtatsch:
Thanks for catching that. Could have sworn I changed that already.

PR #1195, opened 2019-10-15
OpenRTSP to record RTSP H264 stream as video

cadavre:
*This is an all-in-one PR. Please merge previous PRs or just this one if all is ok.*
*Can also cherry-pick `57a61a5`.*

In `firmware_mod/scripts/detectionOn.sh:28` we read:

> Use avconv to stitch multiple JPEGs into 1fps video.
> I couldn't get it working another way.
> /dev/videoX inputs fail.
> Localhost rtsp takes very long (10+ seconds) to start streaming and gets flaky when when memory or cpu are pegged.
> This is a clungy method, but works well even at high res, fps, cpu, and memory load!

I've found that openRTSP handles recording localhost RTSP quite well. **Now tested for more than 2 months.** No problems with mem/cpu usage or getting camera flaky.

This PR enables user to enable `$video_use_rtsp = true` so videos recorded on motion detected will use RTSP instead of 1-second getimage stitch.

**It is disabled by default so no BC is introduced.**


jmtatsch:
Thank you for your magnificent work.
Cannot check in the  openRTSP.bin without reviewing it however.
Is it vanilla or were patches needed? Any build recipes?

cadavre:
openRTSP is vanilla, I cannot remember now where I'd got it, but I'm pretty sure it was https://github.com/Dafang-Hacks/live .

FYI http://www.live555.com/liveMedia/

jmtatsch:
Ok I will attempt to build a openRTSP binary then.

cadavre:
PS I ended up with CBR 2000 kbps setting for 1280*720/15 video to be glitch-free and smooth.

cadavre:
I believe this PR can be closed due to #1196 already merged.

Only waiting for binaries.

jmtatsch:
Yes. Can you make a PR for your best Bitrate settings as well?

cadavre:
I think bitrate settings differ depends on the camera model. Also these settings are general RTSP settings, not strictly RTSP-recording matter.

PR #1204, opened 2019-10-21
Update motion.conf.dist

saipsa:
This value is picked up from telegram.conf

I wasted an hour trying to change values in motion.conf

### DIFF
--- a/firmware_mod/config/motion.conf.dist
+++ b/firmware_mod/config/motion.conf.dist
@@ -21,7 +21,6 @@ publish_mqtt_message=false
 publish_mqtt_snapshot=false
 send_email=false
 send_telegram=false
-telegram_alert_type=image
 send_matrix=false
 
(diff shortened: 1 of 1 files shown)


saipsa:
@jmtatsch Ah yes, it is not set by default. But the bot daemon sets it in telegram.conf, and motionOn also reads the value from telegram.conf. 

https://github.com/EliasKotlyar/Xiaomi-Dafang-Hacks/blob/77e05d4818041620ec1c5d70abd675bf25402b95/firmware_mod/scripts/telegram-bot-daemon.sh#L34



the value in motion.conf is relic from old times, not the other way.

jmtatsch:
Actually the telegram_alert_type belongs to the motion configuration.
Fixed in 87ab7953315a1d541ddd38e67703d47bd744193f.

PR #1217, opened 2019-11-05
update scripts format for one line conditionals

joemulray:
- use cleaner approach for conditionals


saipsa:
Thanks for the PR.

However I am not in favor of it, due to 

1. code consistency. same style for one line and multi line would be better IMO.

2. effort needed to change one line conditional to multiline, if needed in future.

jmtatsch:
3. somewhat reduces readability for me and i do not really see performance improvements through this

Thank you for helping to improve the dafang hacks though.

PR #1220, opened 2019-11-10
Support authenticated RTSP for record_video

nanoamp:
Using openrtsp (since #1195) doesn't currently work with authenticated RTSP streams. If you set `$video_use_rtsp = true` when RTSP requires credentials, then it doesn't record the stream. (It does throw an auth error, but this is normally invisible.)

This PR adds basic support for passing through credentials. You have to manually add them to _motion.conf_, duplicating what's in _rtspserver.conf_ but, as far as I can tell, using openrtsp currently requires manual edits to that conf file anyway.

### DIFF
--- a/firmware_mod/config/motion.conf.dist
+++ b/firmware_mod/config/motion.conf.dist
@@ -35,6 +35,9 @@ video_use_rtsp=true
 video_rtsp_w=1280
 video_rtsp_h=720
 video_rtsp_f=15
+# Uncomment these with appropriate values from rtspserver.conf if set there
+#video_rtsp_username="user1"
+#video_rtsp_password="password2"
 
 # Snapshots
 save_snapshot=false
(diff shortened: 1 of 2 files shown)


jmtatsch:
Shouldn't require manual edits in motion.conf to use openrtsp.
Why not simply sourcing rtspserver.conf in the motion detection script?

nanoamp:
> Shouldn't require manual edits in motion.conf to use openrtsp.

Oh, I couldn't see any way to turn on openrtsp recording without changing `video_use_rtsp` to `true` in an editor. If I'm missing some GUI somewhere, that'd be useful to know.

> Why not simply sourcing rtspserver.conf in the motion detection script?

There already seem to be params in _motion.conf_ which duplicate those in _rtspserver.conf_ (eg. `video_rtsp_w` and `_h`) so I was just following that pattern. If there's no risk of any shell variables in the server config conflicting, then it'd certainly be simpler to just source that in instead.

PR #1239, opened 2019-12-30
Add 768x432 video resolution

Knurlnheim:
Hello,

I needed a lower resolution with 16/9 aspect ratio, and with both dimension divisible by 8 for Motioneye.
This is what I ended up with.

Regards !


jmtatsch:
Thanks for contributing.

PR #1244, opened 2020-01-07
Update tinycam.md to include official instructions

cwiggs:
Since tinyCam version 10.2 there is a Xiaomi Dafang camera that can be used, therefore eliminating the need for custom_vendors.xml.

I left the custom_vendors.xml instructions in case someone needs it, but I'm not opposed to removing it.


jmtatsch:
Thank you for contributing.

Issue #1145, opened 2019-09-01
SW Night Mode only working if Page is open


BMOtheConsole:
On my Dafang Pan/Tilt Auto Night Mode is not working correctly, it is working flawless if im in the SW Night Configuration Tab on my Camera, but as soon as i leave the Page the SW Night Mode is not working anymore and the Camera is Stuck in Whatever Mode it was last

gerardoaguilar:
I am also facing this issue on my two Wyze V2 cameras.

saipsa:
the process is being killed it seems. see what process is started when the page is opened and add it to chron job for every 15 minutes.

danmed:
Also got this problem.. any actual solutions?

dixnor:
The only way I found and seems rock solid is to move the /system/sdcard/www/sw_night_configure.html to another location after you had configured the sensor thresholds.
This just to make sure that you never go back again to nightmode software configuration.
It seems that this html page interfers with autonightmode process in a certain way after you hit the page. Probably because it locks the process when reading the light sensor values.
Give it a try, works

mbaruzzo:
I have a dafang camera and experiencing the same issue, tried dixnor method but no luck. 
Not sure what else to try...

davidpesce:
Same issue here.

msmarks:
I run /system/sdcard/controlscripts/auto-night-detection in ssh, then it works, i don‘t know why.

TRSx80:
> I run /system/sdcard/controlscripts/auto-night-detection in ssh, then it works, i don‘t know why.

I did this method and it works.

That seems to be a script that (with no options supplied (it can also do some other things)) checks if auto-night-detection is already running, and if not, starts it. It also looks in a `CONF_FILE` which by default is `/system/sdcard/config/autonight.conf` and in my case contains:
`-S -j 3 -w 3 -1 1200000 -2 930000,14,10 -3 3000,17,8`.

Those settings look pretty familiar. I think that once you tweak them in the SW detection web page, you press the save button on that page, and it dumps them to this file. Then they can be invoked this way.

Only question remaining is, do we execute this just once each time the camera starts, or periodically with a cron job to check the conditions? I guess I will find out when it gets dark whether it adjusts automatically or not...

vdwkes:
Is there a possibility that auto night detection is always on , on startup of the camera?
[...]

PR #1245, opened 2020-01-07
create autonight.conf to fix auto night mode issue

cwiggs:
This PR was created as a result of this issue: https://github.com/EliasKotlyar/Xiaomi-Dafang-Hacks/issues/1145

Adding this file with the same default settings that are in the "SW night configuration" page fixes an issue where even if
auto night mode is enabled it does not seem to work.  

This also fixes an error in `/var/log/lighttpd-cgi-stderr.log` when auto night mode is enabled, the error is `cat: can't open '/system/sdcard/config/autonight.conf': No such file or directory`

If anyone has suggestions on what these settings do I'd love to add them to the file.


jmtatsch:
At https://github.com/EliasKotlyar/Xiaomi-Dafang-Hacks/blob/77e3ac8b7c049f543e7a1f515e0be9aa7c0655e5/firmware_mod/www/cgi-bin/state.cgi#L44 this will make cameras with superior hw light sensor use the inferior sw auto night path.

cwiggs:
Hm okay, sounds like it might be best to update that file to check for something different.

In the System > Video > Auto Night Mode there is a radio button for HW/SW, perhaps we can use that same method in the cgi file?  I'm not sure where those options come from though, any ideas?

cwiggs:
Perhaps it would better to just add documentation on how to enable the SW auto night setting?

Initially it looks like when you go to the "SW night configuration" page in the gui the SW night config is enabled, but really that page is making calls to the sw night application, using the "config" or "Commandline to use".  However if you want it to persist even when you aren't on the page you have to click the "save" button and stop/start the auto night mode.

Personally I'd rather have a button that allows you to switch between SW and HW mode, although I think the radio buttons under System > Video just tell you which mode is active, it doesn't allow you to control the mode.

@jmtatsch what do you think?

jmtatsch:
Sorry I didn't get back to you for quite some time.
After reevaluating this whole mess, i would suggest the following:

1) autonight.conf should exist by default and look like this:
AUTONIGHT_MODE="HW"
HW_PARAMETERS=""
SW_PARAMETERS="-S -j 3 -w 3 -1 1200000 -2 930000,14,10 -3 3000,17,8"

2) https://github.com/EliasKotlyar/Xiaomi-Dafang-Hacks/blob/88cf55cce1126634effb99d0698bde6c2a378192/firmware_mod/www/cgi-bin/action.cgi#L421 and the following functions have to be changed to toggle HW->SW->HW in autonight.conf

3) https://github.com/EliasKotlyar/Xiaomi-Dafang-Hacks/blob/66d42918df1bbd1f2fedd7b08fb9c6aa879e02fc/firmware_mod/www/cgi-bin/system_video.cgi#L214 has to be changed to read from a function like the next one

4) https://github.com/EliasKotlyar/Xiaomi-Dafang-Hacks/blob/88cf55cce1126634effb99d0698bde6c2a378192/firmware_mod/www/cgi-bin/state.cgi#L43 needs to be adjusted

5) UI in https://github.com/EliasKotlyar/Xiaomi-Dafang-Hacks/blob/66d42918df1bbd1f2fedd7b08fb9c6aa879e02fc/firmware_mod/www/cgi-bin/system_video.cgi#L214 must be adjusted

6) saver for software autonight must be adjusted

Could you please rework your pull request accordingly?

Issue #1241, opened 2020-01-02
Feature: trigger record service via MQTT


apat183:
### Description
Feature request, just like you can turn on motion detection and bunch of other services via MQTT, would be awesome to expose the record service and have that as a MQTT switch. Happy to try and contribute but looking at the mqtt-control.sh I couldn't see and understand how it calls other services to turn on etc.

jmtatsch:
Good idea, you can use https://github.com/EliasKotlyar/Xiaomi-Dafang-Hacks/blob/143aba1159a5aae2acb35235cb48049fd95043b4/firmware_mod/scripts/mqtt-control.sh#L329 as a blueprint. Call "recording on" or "recording off" from common functions.

apat183:
Submitted pull request, can close for now :) https://github.com/EliasKotlyar/Xiaomi-Dafang-Hacks/pull/1246

PR #1246, opened 2020-01-08
Added ability to call recording function via mqtt

apat183:
As per feature request here: https://github.com/EliasKotlyar/Xiaomi-Dafang-Hacks/issues/1241, added the added mqtt switch to call recording function


jmtatsch:
Great addition. Thanks!

PR #1251, opened 2020-01-22
Allow motion recording on protected rtsp link

aeckard87:
Added rtsp config var's USERNAME and PASS, if set, to `detectionon.sh` on record section of script to allow protected rtsp's to record. This does assume that both username and password are set and not one over the other.

Previously recording assumed no authentication on local, rtsp still required it if authentication was set. 

Also added rtsp config var PORT to `detectionon.sh` to allow users to set desired port for rtsp in UI rather than hard coding.


jmtatsch:
Thanks

aeckard87:
You're so welcome @jmtatsch - really appreciate such an active community and responsive owner! Glad to have contributed a little part.

PR #1252, opened 2020-01-23
Multiple Xiaofang FW versions

ZipperZ:
Before installing the "hack" I went through the update process on the original fw and on each step dumped the fw for backup.
138 is still "allowed" to be used not in mainland china.
197 is banned and in serial displays "this device is illegal" and on app, this device is only to be used in mainland china, but still the hack install worked on this version.

The FW was dumped in two ways, one through the serial, another by desoldering the flash chip and reading it with the el-cheapo programming device.


jmtatsch:
Thanks!

PR #1253, opened 2020-01-24
** DO NOT MERGE BEFORE TESTING **

nik0:
- New binaries (latest version of webserver, ftpserver, openssl, ...)
- Latest (I could found of drivers and lib IMP)
- Latest Live555 libs used to compile and latest v4l2rtspserver)
- There is a new way of detecting chipset, I tested with jxf22 but NOT
with jxf23. Please report logs with jxf23
- Not sure audio is working anymore
- Corrected issue: no motion detection at bootup

- Cherry on the cake: better new RTSP server (faster, more stable). Please test it (same problem with jxf23 chipset)
 Less startup options
OSD is flipping for now, need to check ....


jmtatsch:
Amazing! Welcome back :)

fredless:
still looking for testers, or find something evil?

I have some jxf23 wyzev2's I can drop this on if you're still looking for feedback.

nik0:
It would be nice if you could test the new binary and report the traces. I am not sure about the jxf23 chipset detection (the detection driver changed)
Thanks !
I think we need to create a beta test branch, with an option in autoupdate ...

PR #1269, opened 2020-02-08
New menu organisation

gysmo38:
Hello,

I organise menu in three items: configuration, system and informations.
I think  it is more logical.

Gysmo


jmtatsch:
Much better. Thanks.

PR #1271, opened 2020-02-10
Mqtt autodiscovery

gysmo38:
I update mqtt_discovery:
- change icon to be more logical
- change device name from Dafang to DEVICE_NAME in mqtt.conf
- add some variables to avoid code repetition


gysmo38:
I will make a new PR

PR #1283, opened 2020-02-18
Add extra status code to rtsp-h264

hank:
I created a cron watchdog script and when the pidfile was missing status() was returning 0. I made it return 2 in this case. This makes my script work as intended.

### DIFF
--- a/firmware_mod/controlscripts/rtsp-h264
+++ b/firmware_mod/controlscripts/rtsp-h264
@@ -24,6 +24,8 @@ status()
   if [ "$pid" ]; then
     # Prints PID: $pid if exists and returns 0(no error) else returns 1(error condition)
     kill -0 "$pid" >/dev/null && echo "PID: $pid" || return 1
+  else
+    return 2
   fi
 }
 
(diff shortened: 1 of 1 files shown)


saipsa:
is this new code consistent with error codes returned else where? 

If it is new error code, it should be documented somewhere prominent, so that folks adding similar error codes in future know about it.

jmtatsch:
Also I don't understand how exactly this should help. Nothing is expecting an error code 2, is it?
Can you please elaborate?

hank:
The script that my from job runs expects it. I found a case where the rtsp server wasn't running, and there was no pidfile, but status returned 0, which indicates it *is* running. 1 is returned in the case that the pidfile exists and the pid is dead. 2 is standard for file not found in multiple oses, so I didn't think adding anywhere was necessary.

hank:
If it makes it more clear we can make it return ENOENT.

jmtatsch:
Can you also post your watchdog script, might be helpful for others as well.

hank:
```bash
#!/bin/sh
/system/sdcard/controlscripts/rtsp-h264 status || /system/sdcard/controlscripts/rtsp-h264 start
```

jmtatsch:
And you run this as cron job every x minutes?

hank:
> And you run this as cron job every x minutes?

15m yes.

PR #1287, opened 2020-02-22
Fix recording script when RTSP port has been changed from default

benru:
The recording script currently fails if you have changed the RTSP port from the default of 8554. This fix uses the variable PORT which is set in rtspserver.conf to build the correct URL for avconv.

### DIFF
--- a/firmware_mod/controlscripts/recording
+++ b/firmware_mod/controlscripts/recording
@@ -1,6 +1,8 @@
 #!/bin/sh
 PIDFILE="/run/recording.pid"
 DCIM_PATH="/system/sdcard/DCIM"
+
+PORT=8554
 if [ -f /system/sdcard/config/rtspserver.conf ]; then
   . /system/sdcard/config/rtspserver.conf
 fi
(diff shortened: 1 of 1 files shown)


jmtatsch:
Thanks.

Combie81:
is the port variable actually required in firmware_mod/controlscripts/recording on line 5 as it's already defined in /system/sdcard/config/rtspserver.conf on line 13?

Shouldn't it pull the value from there as this is where the web interface changes/saves it and having it in both may lead to a conflict or am I understanding this wrongly?

benru:
@Combie81 There shouldn't be a conflict because the variable is set to 8554 and then immediately overwritten by the rtspserver.conf file.

I was being defensive with my change so that the recording script maintained its old behaviour if you don't have an rtspserver.conf file, but perhaps that is unnecessary because you would have bigger problems if that file were missing.

jmtatsch:
@Combie81 Thanks for paying attention. I would also rather pull the config value from rtspserver.conf as a single source of truth.
@benru Thanks for making defensive changes. rtspserver.conf should always exist otherwise the user is in much greater trouble as you say.

benru:
@jmtatsch @Combie81 I have removed the lines but had to raise another pull request as this one is closed. Thank you both for the review!

PR #1292, opened 2020-02-28
Fix exceeds received frame size

TWijit:
Fix "The total received frame size exceeds the client's buffer size" when record RTSP video.


TWijit:
It result in corrupted video
[https://drive.google.com/folderview?id=1qdmQmikTHCZZnIPXwACpD3Yo2UcHf_Kd](url)
Play this video in windows result in bsod.
My camera is xiaofang 1S. Video record at 720p 15 fps.

jmtatsch:
Ah ok, for me the bottom of your video is green/corrupted. I've had this issue as well and just reduced the resolution until it went away :( Your solution is obviously much superior :)

PR #1298, opened 2020-03-01
Motionvideo2mqtt

Knurlnheim:
This is to add options to push motion captured video through MQTT.


jmtatsch:
How do you use that? Isn’t the message size far outside the mqtt spec then?

Knurlnheim:
Honestly, I didn't check the mqtt spec. I tried it, and it worked... 
Perhaps `mosquitto_pub` is clever enough to send it small pieces by small pieces.

I use that with a mosquitto server instance and node-red. It allows me to centralize things for the house, and not have the cameras able to talk to the internet (and vice-versa).

It works for me, I though I'd share back.

jmtatsch:
Hmm, seems I was mistaken, max message size seems to be 256 mb http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718023. Thanks for you pull request.

PR #1306, opened 2020-03-06
little touch-up for telegram-bot-daemon.sh

FaustoMoon:
My bot sometimes receives a non-message json (spam I think) like this:

{"ok":true,"result":[{"update_id":271754855, "channel_post":{"message_id":73,"chat" ......

in that case updateId has a value and chatId is empty so the markAsRead is never performed.

In order to avoid that some non-message confusing the bot I propose insert an additional check.


jmtatsch:
Thanks

PR #1307, opened 2020-03-07
Disabling SSL / path corrections / misc changes

Combie81:
Lighttpd.conf.dist: Added notes for disabling SSL
SSL left as on by default
Motion.conf.dist: Change pattern to make more sense when items are listed in explorer
Change paths case
resolv.cont.dist: Added public nameservers so that NTP will work by default if a static IP is set by user
rtspserver.conf.dist: Added missing settings that are created after UI page is saved anyway
sendmail.conf.dist: Clarify AUTH entry to match rest of example
recording: Add Recording to path to help seperate/categorise files
Add sub dir with date for faster dir loading
Change from MOV to REC to identify video as "recorting" type
detectionON.sh: Change group/file pattern to make list better/consistent
telegram-bot...: Remove blank line
timelapse.sh: Change path variable name to match rest of code
Add sub directories to help with loaading/filtering of files

### DIFF
--- a/firmware_mod/config/resolv.conf.dist
+++ b/firmware_mod/config/resolv.conf.dist
@@ -9,7 +9,8 @@
 ## List DNS server addresses, primary, secondary, etc
 #nameserver 192.168.1.1
 #nameserver 192.168.1.2
-#nameserver 192.168.1.3
+nameserver 1.1.1.1
+nameserver 8.8.8.8
 
 ## Domain of this system
 ## Determined automatically from fully-qualified hostname when unspecified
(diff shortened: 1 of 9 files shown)


firmware_mod/config/resolv.conf.dist
```diff
@@ -9,7 +9,8 @@
 ## List DNS server addresses, primary, secondary, etc
 #nameserver 192.168.1.1
 #nameserver 192.168.1.2
-#nameserver 192.168.1.3
+nameserver 1.1.1.1
```
jmtatsch: Not a fan of default dns servers, especially not cloudflare and google.
The average joe gets his dns server via dhcp.
The advanced users that want to setup fixed ip adresses also have to set their own dns servers.
We can keep your suggested dns servers commented out as mere suggestion.

Combie81: Commit updated - I thought this would only take effect though if a static address is assigned - but having them as suggestions works too :)

PR #1310, opened 2020-03-12
Fix #1309 auth on http

### LINKED ISSUE #1309: Lighttpd: user/password sent in clear (HTTP) before redirect to HTTPS
Hello,

I noticed that my browser is asking for username/password two times when connecting to the camera from browser coldstart.
In lighttpd.conf, we ask for authentification even if protocol is http, resulting in cleartext transmission... and only then we redirect to https.

I think the fix is simple, I will try to push a PR today.

Knurlnheim:
This fixes #1309.
We only require auth on https; if on plain http, we redirect (to https).

Warning : comments to remove SSL by commenting "server :443" and  "redirect" directives block will end up with http only and no authentication.. But this is probably a good thing, since http + auth is very very bad as it provides a false sense of security...


jmtatsch:
Thanks.

Issue #1301, opened 2020-03-02
FIX INCLUDED: SW Nightmode not working


0n3-70uch:
* exact camera hardware: Xiaomi Dafang
* also confirmed: Neos SmartCam (https://github.com/EliasKotlyar/Xiaomi-Dafang-Hacks/issues/1297)
* the Xiaomi-Dafang-Hacks commit id: d9fde69 (latest)

### Description

Do not auto switch to night or day mode.

Everything similar to this:
https://github.com/EliasKotlyar/Xiaomi-Dafang-Hacks/issues/1300

### What did you do to debug the issue

See link above.

### Evidence

<details> 
<summary>Summary of all log files</summary>
<div>
<pre>
==> /var/log/isp_tuning_func <==
0x0000000f
�?wl�C0�whdw
==> /var/log/lighttpd-access.log <==
192.168.2.11 192.168.2.13 root [02/Mar/2020:17:10:22 +0100] "GET /cgi-bin/currentpic.cgi HTTP/1.1" 200 334512 "-" "HomeAssistant/0.106.2 aiohttp/3.6.1 Python/3.7"
192.168.2.11 192.168.2.13 root [02/Mar/2020:17:10:32 +0100] "GET /cgi-bin/currentpic.cgi HTTP/1.1" 200 334503 "-" "HomeAssistant/0.106.2 aiohttp/3.6.1 Python/3.7"
192.168.2.11 192.168.2.13 root [02/Mar/2020:17:10:42 +0100] "GET /cgi-bin/currentpic.cgi HTTP/1.1" 200 334356 "-" "HomeAssistant/0.106.2 aiohttp/3.6.1 Python/3.7"
192.168.2.20 192.168.2.13 root [02/Mar/2020:17:10:44 +0100] "POST /cgi-bin/action.cgi?cmd=showlog HTTP/1.1" 200 1251 "https://192.168.2.13/" "Mozilla/5.0 (X11; Fedora; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/80.0.3987.122 Safari/537.36"
192.168.2.11 192.168.2.13 root [02/Mar/2020:17:10:52 +0100] "GET /cgi-bin/currentpic.cgi HTTP/1.1" 200 334183 "-" "HomeAssistant/0.106.2 aiohttp/3.6.1 Python/3.7"
[...]

PR #1312, opened 2020-03-15
Save correct SW auto night command line arguments

fhl206:
autonight.conf.dist has arguments enclosed with ' but SW night configuration page passes arguments without. Currently, updating autonight.conf sw parameters using the SW night configuration page will stop autonight working correctly.


jmtatsch:
Thanks for contributing

jmtatsch:
Fixes #1301

PR #1313, opened 2020-03-17
request snapshot over mqtt

josephdouce:
${TOPIC}/motion/mqtt_snapshot GET

the above command returns a snapshot compressed to 25Kb

### DIFF
--- a/firmware_mod/scripts/mqtt-control.sh
+++ b/firmware_mod/scripts/mqtt-control.sh
@@ -241,10 +241,10 @@ done
       /system/sdcard/bin/mosquitto_pub.bin -h "$HOST" -p "$PORT" -u "$USER" -P "$PASS" -t "${TOPIC}"/motion/mqtt_snapshot ${MOSQUITTOPUBOPTS} ${MOSQUITTOOPTS} -m "$(motion_mqtt_snapshot status)"
     ;;
         
-    "${TOPIC}/motion/mqtt_snapshot GET")
+    "${TOPIC}/mqtt_snapshot GET")
       /system/sdcard/bin/getimage > "/tmp/mqtt_snapshot"
       /system/sdcard/bin/jpegoptim -S 25k "/tmp/mqtt_snapshot"
-      /system/sdcard/bin/mosquitto_pub.bin -h "$HOST" -p "$PORT" -u "$USER" -P "$PASS" -t "${TOPIC}"/motion/snapshot/image ${MOSQUITTOPUBOPTS} ${MOSQUITTOOPTS} -f "/tmp/mqtt_snapshot"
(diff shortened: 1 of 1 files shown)


firmware_mod/scripts/mqtt-control.sh
```diff
@@ -240,6 +240,13 @@ done
     "${TOPIC}/motion/mqtt_snapshot")
       /system/sdcard/bin/mosquitto_pub.bin -h "$HOST" -p "$PORT" -u "$USER" -P "$PASS" -t "${TOPIC}"/motion/mqtt_snapshot ${MOSQUITTOPUBOPTS} ${MOSQUITTOOPTS} -m "$(motion_mqtt_snapshot status)"
     ;;
+        
+    "${TOPIC}/motion/mqtt_snapshot GET")
+      /system/sdcard/bin/getimage > "/tmp/mqtt_snapshot"
+      /system/sdcard/bin/jpegoptim -S 25k "/tmp/mqtt_snapshot"
+      /system/sdcard/bin/mosquitto_pub.bin -h "$HOST" -p "$PORT" -u "$USER" -P "$PASS" -t "${TOPIC}"/motion/snapshot/image ${MOSQUITTOPUBOPTS} ${MOSQUITTOOPTS} -f "/tmp/mqtt_snapshot"
```
jmtatsch: This can potentially lead to mixups for users that expect only motion sensed events on this topic, how about just /snapshot/image?

josephdouce: snapshot/image work's I'll update and re-submit

firmware_mod/scripts/mqtt-control.sh
```diff
@@ -240,6 +240,13 @@ done
     "${TOPIC}/motion/mqtt_snapshot")
       /system/sdcard/bin/mosquitto_pub.bin -h "$HOST" -p "$PORT" -u "$USER" -P "$PASS" -t "${TOPIC}"/motion/mqtt_snapshot ${MOSQUITTOPUBOPTS} ${MOSQUITTOOPTS} -m "$(motion_mqtt_snapshot status)"
     ;;
+        
+    "${TOPIC}/motion/mqtt_snapshot GET")
```
jmtatsch: This doesn't have to do with motion.
Also GET is somewhat inconsistent with our existing style. 
How about /snapshot/take_image

josephdouce: GET was to clarify this is a request that responds only via mqtt /snapshot/take_image may be misinterpreted as saving the snapshot to file as well as mqtt

jmtatsch:
Thanks for contributing

PR #1323, opened 2020-03-28
enable zram by default as per the original firmware

Yannik:
this prevents crashes due to memory exhaustion on xiaofeng and other
64mb models

### DIFF
--- a/firmware_mod/config/swap.conf.dist
+++ b/firmware_mod/config/swap.conf.dist
@@ -2,6 +2,7 @@
 # in particular on the xiaofang & clones with only 64 mb 
 
 SWAP_ZRAM=true
+SWAP_ZRAM_SIZE=16777216
 
 # in case you want to use an additional swapfile, enable it here
 SWAP=false
--- a/firmware_mod/config/swap.conf.dist
+++ b/firmware_mod/config/swap.conf.dist
@@ -3,6 +3,7 @@
 
 SWAP_ZRAM=true
 
-SWAPFILE=false
-SWAPFILE_PATH="/system/sdcard/swapfile"
-SWAPFILE_SIZE=256
+# in case you want to use an additional swapfile, enable it here
+SWAP=false
+SWAPPATH="/system/sdcard/swapfile"
+SWAPSIZE=256
--- a/firmware_mod/config/swap.conf.dist
+++ b/firmware_mod/config/swap.conf.dist
@@ -1,6 +1,8 @@
 # If you need to run many services in parallel you may want to enable swap on your camera to avoid running out of memory
 # in particular on the xiaofang & clones with only 64 mb 
 
-SWAP=false
-SWAPPATH="/system/sdcard/swapfile"
-SWAPSIZE=256
+SWAP_ZRAM=true
+
+SWAPFILE=false
+SWAPFILE_PATH="/system/sdcard/swapfile"
+SWAPFILE_SIZE=256
--- a/firmware_mod/run.sh
+++ b/firmware_mod/run.sh
@@ -68,7 +68,7 @@ if [ "$SWAP" = true ]; then
     echo "Swap file created in $SWAPPATH" >> $LOGPATH
   fi
   echo "Configuring swap file" >> $LOGPATH
-  swapon $SWAPPATH
+  swapon -p 10 $SWAPPATH
   echo "Swap set on file $SWAPPATH" >> $LOGPATH
 fi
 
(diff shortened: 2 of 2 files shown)


jmtatsch:
This is a great find but the implementation will cause a lot of issues with people's existing swap configurations.
Could you please refactor to keep the old config names and just enable the zram case if
SWAP=true
USE_ZRAM=true

Yannik:
While refactoring this, I wanted to write some documentation about the feature but noticed that there are a few things I don't quite understand yet.
First, I thought the zram would be the ram described as `usable after init` in the kernel log
```
[    0.000000] Determined physical RAM map:
[    0.000000]  memory: 0051b000 @ 00010000 (usable)
[    0.000000]  memory: 00035000 @ 0052b000 (usable after init)
[    0.000000] User-defined physical RAM map:
[    0.000000]  memory: 028b9000 @ 00000000 (usable)
```

However, that is not quite true. Per my research, zram makes use of the 35404KB (on my xiaofeng) of ram listed in /proc/meminfo as `MemTotal`, but compresses the data before storing it.

There's also the `rmem` and `ispmem` specified as kernel boot option:
```
console=ttyS1,115200n8 mem=41700K@0x0 ispmem=8M@0x28B9000 rmem=15644K@0x30B9000
```

However, I don't quite understand why

1. there's only 35404KB `TotalMem` even though 41700K are specified
2. what is the ispmem actually used for?
3. what is the rmem actually used for?
4. what's up with the memory region marked `usable after init`  in the boot log?

As it stands, enabling zram is enabled by default on stock firmware, and it also fixed by problems of out-of-memory crashes, so this patch should definitely go forward.
However, I'd also like to understand this a bit more, as I could not find any documentation on these parameters on the internet. :-)

jmtatsch:
@EliasKotlyar Can you help?

Yannik:
>     1. there's only 35404KB `TotalMem` even though 41700K are specified
Some space is taken by the kernel:
```
Memory: 35192k/41700k available (3951k kernel code, 6508k reserved, 1276k data, 212k init, 0k highmem)
```
I think the 212k init make up the difference to the `35404KB` total

>     4. what's up with the memory region marked `usable after init`  in the boot log?
The (automatically) "determined physical RAM map" is never used because we specify a user-defiend physical RAM map:

```
[    0.000000] Determined physical RAM map:  <--- this is automatically determined
[    0.000000]  memory: 0051b000 @ 00010000 (usable)
[    0.000000]  memory: 00035000 @ 0052b000 (usable after init)
[    0.000000] User-defined physical RAM map: <- this overwrites the automatically determined memory maps
[    0.000000]  memory: 028b9000 @ 00000000 (usable)
```

Yannik:
>     2. what is the ispmem actually used for? 
>     3. what is the rmem actually used for?

According to https://github.com/Dafang-Hacks/Ingenic-T10_20/blob/master/doc.7z:


> 3. 预留内存及内核裁减
> 
> 预留内存分为两部分：ispmem及rmem：
> 
> * ispmem为ISP所需内存，计算方法为Sensor输出图像长x宽x4，例如对于720P的Sensor为1280x720x4=3600KB，对于960P的Sensor为1280x960x4=4800KB。若需要打开WDR功能，ispmem需要2倍的大小（即长x宽x4x2）。
> * rmem为系统Video Buffer所需内存，720P大概需要14MB内存，960P大概需要19M内存。

which roughly translates to:

> 3. Reserved memory
> 
> The reserved memory is divided into two parts: ispmem and rmem:
> 
> * ispmem is the memory required by the ISP, and the calculation method is Sensor output image length x width x4, for example, 1280x720x4 = 3600KB for 720P Sensor, 1280x960x4 = 4800KB for 960P Sensor. If you need to turn on the WDR function, ispmem needs twice the size (ie length x width x4x2).
> * rmem is the memory required by the system Video Buffer, 720P requires 14MB memory, 960P requires 19M memory.

`WDR` means Wide Dynamic Range (like HDR), which we are not using. So according to this, `8100KB` (=7,91MB) ispmem would be necessary for 1080p.
`isp` means the `Image Signal Processor` of the T20:
image


I had a really hard time finding out where these cmdline parameters are actually used in the kernel. I see no reason not to trust these values in the docs, but I'd like to know how this really works.

In `arch/mips/xburst/soc-t20/common/platform.c`:,  `__setup("ispmem=", ispmem_parse);` sets `ispmem_size` and `ispmem_base` variables, but these are not used anywhere else in the kernel.

I couldn't find any place where `rmem=` is parsed in the kernel. However, I can confirm that v4l2 rtspserver fails to start when all memory is used for user-space memory and ispmem. There are no references to `rmem` in the v4l2 code either though.

So if @EliasKotlyar has more information on this, I'd be glad to learn.

Yannik:
I finally found out more about `rmem`.

[rmem.c](https://github.com/Dafang-Hacks/kernel/blob/master/drivers/misc/rmem.c) creates a `dev/rmem` device which is an exclusive memory region to be used by `libimp`. 

Use `pmap $(pidof v4l2rtspserver-master) |grep rmem` to examine the `rmem` memory usage.

Yannik:
@jmtatsch I updated my PR as per your request.

PR #1333, opened 2020-04-05
New binaires (latest internet sources)

nik0:
bftpd
	busybox
	fatlabel
	fsck.fat
	lighttpd
	mkfs.fat
	mosquitto
	openssl


jmtatsch:
Are these binaries somewhat tested to work with our current functionalities?

nik0:
Yes,
I will add them to the beta branch now. I am just wondering how to do if I want to publish them separately (to be tested one by one)

jmtatsch:
I think all at once on the beta branch should be fine.

Yannik:
@nik0 Which sources are used for these binaries exactly?

nik0:
See the source tree here: https://github.com/Dafang-Hacks/Main 
in "Modules" folder

tachang:
@nik0 Is avconv one of those sources?

nik0:
> @nik0 Is avconv one of those sources?

No sorry, I did not recompile avconv. Do you need a newer version ?

PR #1348, opened 2020-04-11
Dropbear new version 

nik0:
V2019.78


jmtatsch:
Wonderful, this one works.
Only it still cant find the scp when transferring from outside.
Did you push the updated sources?

PR #1366, opened 2020-04-16
PTZpresets.sh should not use inverted logic for motor control.

slm4996:
PTZpresets.sh moves to an arbitrary location not up or down, left or right. Inverted logic on the motor script breaks this if you flip your video.

### DIFF
--- a/firmware_mod/scripts/PTZpresets.sh
+++ b/firmware_mod/scripts/PTZpresets.sh
@@ -7,7 +7,7 @@
 
 source /system/sdcard/scripts/common_functions.sh
 
-MOTOR=/system/sdcard/bin/motor
+MOTOR=/system/sdcard/bin/motor.bin
 pid=$$
 echo $pid > /run/PTZ_$pid.pid
 


jmtatsch:
I dont get it, please elaborate what the actual error is. And why you propose to bypass the functionality of the flipped modus to make the flipped modus work?

slm4996:
@jmtatsch turn on flipped video "setconf -k f -v 1".
Go to a preset position, "PTVpresets.sh 1600 400". 
Repeat the last command. You'll end up at 1600 0.  The problem is that when PTVpresets wants to go -200 to get towards 1600 400 motor flipping sends it +200 instead.

Basically, a preset targets a fixed position or number of steps, not a direction. 1600 400 is always the same position, it doesn't matter if video is flipped.

jmtatsch:
Makes sense. Can you please check if motion tracking i.e. detectionTracking.sh still works with your proposed change?

slm4996:
It does work. If video is flipped it only works with the change unless you never happen to track motion vertically.

jmtatsch:
Thank you.

PR #1370, opened 2020-04-18
Motion detection was not working if reduce pool is off

nik0:


jmtatsch:
Motion detection seems to work for me again. Thank you.

PR #1371, opened 2020-04-18
Allow example userscripts, with calibratePTZ example script.

slm4996:
Changes run.sh to only execute files in /system/sdcard/config/userscripts/startup/ if they do not end with .dist. This allows example scripts to be included in the repository and/or scripts to be deactivated.

This also includes an example script calibratPTZ.sh.dist that will perform a PTZ calibration at boot and if a /system/sdcard/config/cameraposition file exists it will also move the camera to the position it defines.

The script is disabled by default for two reasons:
1. Not all cameras are PTZ.
2. Cameras that do not have physical endstops can be damaged from repeated calibrations. There is a note about this in the example script.

To enable the calibratePTZ.sh.dist demo, or any other script, copy or rename it to NOT end in .dist.
Example:
cp /system/sdcard/config/userscripts/startup/calibratePTZ.sh.dist /system/sdcard/config/userscripts/startup/calibratePTZ.sh

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -57,5 +57,8 @@ firmware_mod/config/rtspserver.conf
 firmware_mod/config/ntp_srv.conf
 firmware_mod/config/lighttpd.user
 firmware_mod/config/lighttpd.pem
-firmware_mod/config/userscripts/*/*
+firmware_mod/config/userscripts/motiondetection/*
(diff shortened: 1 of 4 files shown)


slm4996:
Sorry for the messy commits, from my tree, it has been a while since I used git. No extra content was changed. If you prefer I can cleanup and send a pull from a feature branch for you.

PR #1381, opened 2020-04-25
Add scp 

Knurlnheim:
Hello,

this little PR adds SCP to and from the camera. I tried to minimize the memory footprint on my Dafangs, but could not test on other models.


Knurlnheim:
Ah, I just figured out that a new dropbearmulti is in the beta branch.
I was confused because in master the symlinks were there but it wasn't working.
Sorry about that, I will close this.

PR #1383, opened 2020-04-26
Update openRTSP according to last version of Live555 libs

nik0:


jmtatsch:
Seems to create non zero size recordings for me. Thank you.

PR #1396, opened 2020-04-30
Start of a unified movement control system.

slm4996:
Anybody else tired of having PTZ/movement control split across multiple files and scripts? I am!
This is the start of consolidating everything into one script that will accept input from cgi, other scripts, etc. It will replace /system/sdcard/bin/motor in handling inverted motion control.

The basic script is done, not other scripts have been updated to make use of it yet.

Wishlist for my ideal motion control system (so far):
- [X ] One central place to make changes to motor control.
- [ ] Min/Max awareness and adjusting user input to not exceed physical limitation of camera movement (do not rely on endstops).
- [ ] Specific point navigation. Enable clicking on camera image and centering that point in view.

**Please do not merge this yet!**
I want to collaborate on all changes. Please comment with your ideas as they compare to mine, any contributions will be welcome! Additional work to update other scripts to make use of this will commence once I have some more feedback on the general direction I have proposed and any community adjustments.

### DIFF
--- a/firmware_mod/scripts/ptz.sh
+++ b/firmware_mod/scripts/ptz.sh
@@ -1,0 +1,300 @@
+#!/bin/sh
+
+###########################################################
+### Begin common paths and variables
+###########################################################
+# Log path (blank will skip logging to file)
+log="/system/sdcard/log/ptz.log"
+
+# Include common_functions from Dafang-Hacks
+# shellcheck disable=SC1091
+. /system/sdcard/scripts/common_functions.sh
+
+# Gets the command name without path
+cmd=$(basename "$0")
+
+# Path to motor binary
+MOTOR=/system/sdcard/bin/motor.bin
+
+# Path to ptz_presets.conf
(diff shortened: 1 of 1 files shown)


jmtatsch:
I completely agree that the current move functionality is quite messy. Imho there should be functionality in the motor binary or deeper to
* track the current position /state ( this State has to survive a restart)
* calibrate properly for cameras with and without endstops (those without having to rely on the aforementioned position/state tracking)
* virtual safety boundaries that cannot be moved over to save cameras without physical endstops

Of course this could all be implemented in scripting but what if both motion tracking is enabled and going somewhere but you want to move somewhere else via onvif. You would need some kind of command queuing. Current onvif implementation blocks the motor driver completely, so no Other movement is possible I think, which is bad.

sedony23:
I think, in the current sate this breaks PTZ control over ONVIF. As PTZ is not working any more in my ONVIF setup after updating. Or how do I fix this?

PR #1397, opened 2020-04-30
Hide dpad if camera have no PTZ

gysmo38:
Check camera model and hide dpad for camera without PTZ.


gysmo38:
I only use the already done  show_HWmode function in action.cgi.
It check if the /driver/8189es.ko or /driver/8189fs.ko is loaded. I don't know if there is others drivers on others dafang.
If you have a list of drivers, I can complete the list. If they use same driver, it should work.

jmtatsch:
You are right.
There is only the wyze cam pan commonly used, but that is detected as dafang.
So it shouldnt be a problem.

PR #1401, opened 2020-05-01
Viewer

gysmo38:
@jmtatsch I wrote a page to show files in Recording directory.  It will resolve link done before and i found it more user friendly.
I rename Recording menu to SD card. I can add a submenu for timelapse.
Another idea is to add a directory for sound and add a page to upload sound.


firmware_mod/www/cgi-bin/action.cgi
```diff
@@ -742,6 +742,26 @@ if [ -n "$F_cmd" ]; then
         fi
         return
         ;;
+     recording_files)
+      for file in $(find /system/sdcard/DCIM/Recording/ -type f)
+        do
+         if [[ -f $file ]]; then
+          file_size=$(ls -lh $file | awk '{print $5}')
+          file_url=$(ls -lh $file | awk '{print $9}' | sed 's/\/system\/sdcard\/DCIM/https:\/\/192.168.76.105\/viewer/')
```
jmtatsch, line 750: ip adress should be more generic :)
you could use this to get your cameras ip adress`ip -o -4 addr show dev eth0 | sed 's/.* inet \([^/]*\).*/\1/'`

firmware_mod/www/cgi-bin/action.cgi
```diff
@@ -742,6 +742,26 @@ if [ -n "$F_cmd" ]; then
         fi
         return
         ;;
+     recording_files)
+      for file in $(find /system/sdcard/DCIM/Recording/ -type f)
```
jmtatsch, line 746: We should list the whole DCIM directory. We could write one *_files script per directory or come up with something more generic.
Maybe have a look at one of these suggesting for listing https://stackoverflow.com/questions/18014779/similar-command-to-linux-tree-command-in-adb-shell

jmtatsch:
Great idea, thanks for working on this.

firmware_mod/www/cgi-bin/action.cgi
jmtatsch: well `ip -o -4 addr show dev wlan0 | sed 's/.* inet \([^/]*\).*/\1/'`

jmtatsch: not sure what to do about them usb ethernet users

gysmo38: I will correct thoses mistakes...

firmware_mod/www/cgi-bin/action.cgi
gysmo38: Yes good idea

gysmo38:
I will work on it today. I will update my PR.

gysmo38:
@jmtatsch I should ok now :) I have others idea for improvement but I will first fix autoupdate.sh

jmtatsch:
Thank you very much. I will transfer everything to the code only repository once the current beta is stable enough to be merged into master.

fanaticsurf:
New menu SD card with sub menu Recordings and Timelapse does nothing in my case. Maybe because I dont use Recordings and Timelapse.
Should motion stills and videos be shown in Recordings or do we need a sub menu Motion?

If I type in IP/viewer in the browser I can still access the motion files.

jmtatsch:
Yes, currently you would need recordings or timelapses to see something.

I get mixed results though. I was able to view timelapses and recordings for a short period of time but now it just does nothing although the files are there. No idea why.

gysmo38:
@fanaticsurf Which directory motion use? I can maka a new PR for motion directory. 
@jmtatsch Do you have a lot of file in recording or timelapse? I juste tested with few files. I will try to add sort and seach in the table.

jmtatsch:
Motion detected stuff goes to DCIM/motion/stills and DCIM/motion/videos
Not really a lot of files, I just enabled timelapse today for testing the viewer.
And in recordings its just 2 test recordings.

if I run www/cgi-bin/list_files.cgi directly: line 47: syntax error: unexpected end of file (expecting "fi") but I dont see a missing fi?

jmtatsch:
Well its over 9000 timelapse images already.
Maybe iterating over find results is not optimal because we have to wait for the search to finish https://stackoverflow.com/questions/9612090/how-to-loop-through-file-names-returned-by-find

Combie81:
If you reload the page it gives you the results multiple times and increases on each reload.  Can you add in a check to clear the list before it searches the files?

gysmo38:
@Combie81  Yes I notice this bug. It will be fixes in next PR.

Combie81:
I forgot to ask if it's also possible to have a delete/mass download/mass delete option? Perhaps checkboxes down the side and option at bottom/top to delete/download selected?

PR #1405, opened 2020-05-03
Add test for update process

gysmo38:
Bug fixes for autoupdate.
We check the branch and version.
If we have already the lastest version, we can switch to other branch or force full update (delete VERSION file and downloas all the firmware again)


jmtatsch:
Can you explain what changes were made to action.cgi? Did you run a reformatter? Soo many changes...

gysmo38:
Not so many changes on action.cgi. I don't known what I did wrong... I just modify update action and remove updatebeta.

gysmo38:
Ok all my action.cgi file is with two space on each line. I will correct this and update the PR

PR #1417, opened 2020-05-08
Bring beta up to date

jmtatsch:


nik0:
Great !!!

jmtatsch:
Haha, sorry for all this mess.

nik0:
It’s not me that will say something 😜

jmtatsch:
Anyhow, thanks a lot for making the rtspserver work great again :)
Did you check in all the changed sources as well?
Maybe the HLS streaming works better now, i will give it a shot...

nik0:
Yes everything is committed...

jmtatsch:
Hmm, 
Flag -S is set
server is sending
send file:index.html
send file:hls.light.min.js
But still no HLS stream

nik0:
How do you check that ?
I tested with VLC and it seemed to work

jmtatsch:
just http://yourip:8554

nik0:
With which browser ?

jmtatsch:
Safari is supposed to have native support. Tried firefox as well.

PR #1418, opened 2020-05-08
Reworking permissions of saved files and directories

### LINKED ISSUE #1413: Wrong permissions on directories created by "save_video=true"
Xiaomi Dafang 128 MB
Commit ec9f5b9 (latest as of today)
Custom bootloader for FullHD recording

I wanted to save recordings on a NFS mount, so I created it and mounted it. Creating files and directories with "touch" or "mkdir" works fine, permissions are as expected (rwx.r.r, rwx.rx.rx for dirs).

I tried two different methods to save motion recordings: custom script and "save_video=true" in motion.conf.

The first option calls the script "recording on" and it works: subfolders are created on the NFS share, the recordings are there. The recordings show garbage in the lower half of the frames but I think it's a known issue.

Using "save_video=true" in motion.conf doesn't work out of the box.

A subdirectory is created with the current day but the permissions are rw.rw.rw, without "x", therefore the recording script itself (I couldn't find it, which one is it?) cannot "enter" the directory, and cannot record files in it.

After I manually change the permissions, the recording works fine until the next day, when a new subdir is created without "x" permission. Also, using this methos I don't get garbage in the lower half of the frames, even at FullHD, so it's my preferred method.

As said, manually creating directories produces the expected "x" permission.
Somehow the "save_video" script generates subdirs in a way that breaks the recording process.

dewi-ny-je:
"save_snapshots_attr" was used also for firectories containing videos, leading to non-obvious permission issues.

Permissions for snapshots and videos were grouped together into a new "save_files_attr" and a new "save_dirs_attr" has been introduced.

Closes #1413

### DIFF
--- a/firmware_mod/config/motion.conf.dist
+++ b/firmware_mod/config/motion.conf.dist
@@ -29,6 +29,8 @@ send_matrix=false
 group_date_pattern="+%Y-%m-%d"
 file_date_pattern="+%Y-%m-%d_%H-%M-%S"
 video_duration=10
+save_files_attr="0644"
+save_dirs_attr="0755"
 
 # Record from RTSP (RTSP H264 must be enabled)
 # RTSP is alternative for `record_video` function in `detectionOn.sh` - instead of 1s-getimage gets real video.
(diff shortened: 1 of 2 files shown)


jmtatsch:
Thanks for helping!

PR #1428, opened 2020-05-13
With this fix the file starts working

domiwar:
With the original file from repository i have blank page, this fix


jmtatsch:
Nice! Thank you!

PR #1441, opened 2020-05-24
[NEW UI] Motion detection page

gysmo38:
Finish motion detection page with multi zones detection, control all important parameters.


gysmo38:
Done in other PR

PR #1446, opened 2020-05-25
[NEWUI] Improve sdcard files

gysmo38:
Put Motion files first
Add delete button
Change Download by fontawesome


gysmo38:
In SD Card page add pagination and search.

PR #1452, opened 2020-05-27
[NEWUI] Bug fixes

gysmo38:
Add type password for all password inputs in motion page
Correct url encodinf in motion page
Add dark-grey theme to try to have a dark theme. Not sure the result will be pleasent :p


gysmo38:
Add accordtion others update in update page to be more clear

Issue #1368, opened 2020-04-17
Telegram Help PLS


ex725:
Hello there
telegram detects commands that I send manually.
But I want a bot to detect the commands it sends.

I made a short video for this
https://youtu.be/kU36mfIEabo

olvier:
Maybe try "/shot@botname"

I remember i wasn't able to manually shot via second-telegram-bot, too (node-red).
But i can't remember whether i tried "/shot" or "/shot@botname".

ex725:
can you explain a little more
I do not understand
![image](https://user-images.githubusercontent.com/56446241/79645148-24d21480-81b6-11ea-8ef2-d08d1c525219.png)

ex725:
> Maybe try "/shot@botname"
> 
> I remember i wasn't able to manually shot via second-telegram-bot, too (node-red).
> But i can't remember whether i tried "/shot" or "/shot@botname".

olvier:
If this didnt work I can't help you. Checked on wife's phone "/on" and "/off" and both work without botname. So I have no idea what's wrong with your setup.

ex725:
?????

ex725:
:(

fergio:
So you have a camera, tablet and phone. That's 3 people in the conversation ? You need group chat and a group chat id on the telegram.conf.

Issue #1364, opened 2020-04-16
Telegram


ex725:
installed in me and my wife ( Telegram )
we created a common group.
but only I can control it.
How do I add multiple chat-id?

ex725:
https://www.youtube.com/watch?v=kU36mfIEabo&feature=youtu.be

jmtatsch:
🤷‍♂️, look at the telegram bot script perhaps?

PR #1453, opened 2020-05-27
Telegram Bot: support `channel_post`

pfefferle:
posts to channels are of the type "channel_post" (not message): https://core.telegram.org/bots/api#getting-updates

This MR supports channel posts and should fix problems like #1368 and #1364

(perhaps there is a simpler solution, I am not that familiar with bash programming)


jmtatsch:
Thanks

s58DB:
How use this option chanel_post?

PR #1466, opened 2020-05-31
Update motion.cgi

Combie81:
Reworked file to add missing settings - will probably require checking by @gysmo38 and the page updated to display the settings


gysmo38:
It look good for me. I do not add thoses parameters because I want to keep interface simple.
Maybe I can try to add a standard and advance view of pages.

Combie81:
An "Advanced" option/view would be good.  Would be nice if all settings were able to be configured from within UI :)

gysmo38:
Yes agree with you :) I will work on it. I  will send you a version of motion cgi with json. Hope it could simplify the code :)
I think I will move all text from cgi to js and just get back the values.

PR #1471, opened 2020-06-01
[SCRIPT] Merging script for config files

gysmo38:
I create a scritp to merge .conf.dist and .conf file.
The script will check if there is new variables on conf.dist file and added in the config file is it exist.

You can test it, if it is ok for everybody we can add this script at the end of the update process. It is to avoid breaking features when there is parameters missing in an old config file.

### DIFF
--- a/firmware_mod/scripts/merge_config.sh
+++ b/firmware_mod/scripts/merge_config.sh
@@ -1,0 +1,28 @@
+#!/bin/sh
+# Check if there is new parameters in the dist config file
+
+CONF_PATH="/system/sdcard/config"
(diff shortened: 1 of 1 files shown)


saipsa:
It would be  better if instead of enumerating all files manually, you scan for all *.conf files and do the merging, so if a new file is added at a later date, it would still work.

gysmo38:
That was my first thought but there is some conf file that have no parameters so it can not be updated.
(example hostname.conf, wpa_supplicant.conf). Or I can change it to create an exclude filter...

jmtatsch:
Exclude filter would be great.

gysmo38:
I will made the correction and made a new PR.

PR #1472, opened 2020-06-02
[NEW UI] Make it default UI 

gysmo38:
What's done:
- rename index2.html to index.html
- remove link to main UI in about box
- move third part library to a new directory lib
- prefix cgi-bin relate to web ui with ui_
- move timezone.tsv to json directory. I will change script to use json instead of tsv. In this new directory I will put json version file and update json file.
- delete old files related to previous UI
- rename all urls and actions to new paths


Combie81:
Bump :P

PR #1481, opened 2020-06-07
[UI] Control service change to toggle

gysmo38:
Change service control to toggle button.
I also re write cgi to be more efficient, for that I need to rename onvif control script to remove .sh on his name.

@jmtatsch Not yet found a way to have the theme color on it. So I let it green.


gysmo38:
I also update rtsp-h264 et rtsp-mjpeg script to change pid filename from v4l2server....pid to rtsp-h264.pid and rtsp-mjpeg.pid
All controlscripts have same way to name pid file.

jmtatsch:
More consistency is always nice. And ui_control.cgi is so much leaner.

jmtatsch:
Hmm, the stati of the services are not read correctly for me.
Screenshot 2020-06-07 at 22 15 37Everything appears stopped although some are undoubtetly running.
Also can you make the play/stop button just one round button that is replaced upon click by its counterpart?

gysmo38:
@jmtatsch strange. I remove buttons start/stop. It is now toggle button. I will check tomorrow what happen.

Combie81:
Yeah I noticed that - i started a fresh install/config and works fine.
Sadly all the spaces are back :P

Issue #1470, opened 2020-06-01
Telegram and mount USB storage


s58DB:
Hello,

yesterday I upgraded FW on SD-card on new version 34a96ba. On previous version I don't have problems with telegram and mount USB storage.

On this version Telegram-bot don't respond on commands when I send them via phone but send image when detect motion. 

With USB storage I don't know what I do wrong. I wrote simple script in /config/userscripts/startup. This Script I make executable but don't run when I reboot system. If i run script manually I get good result.

Does anyone have the same problems?

Best regards, D

pfefferle:
The change only adds an alternate message source, that supports channel posts, it should not break previous implementations.

pfefferle:
Do you send via channel or directly to the bot?

jmtatsch:
Does your userscript have the .sh ending?

s58DB:
> Do you send via channel or directly to the bot?

I send commands directly to the bot. 

> Does your userscript have the .sh ending?

yes

![my_script_mount](https://user-images.githubusercontent.com/23266214/83421982-2cc5db00-a429-11ea-9dfc-cf4f1994d326.PNG)

dvybornov:
I’ve got the same troubles with Telegram-bot after last update. I send messages via shared group with bot.

YogaSakti:
I also have the same problem. when i see the logs page, , last_update_id changes but there is no reply message 
![image](https://user-images.githubusercontent.com/24309806/83999805-aa3f9d00-a98d-11ea-91a1-c9f897aa9b32.png) 
![image](https://user-images.githubusercontent.com/24309806/83999821-b9bee600-a98d-11ea-9981-4279a304cf8e.png)
![image](https://user-images.githubusercontent.com/24309806/83999840-c6433e80-a98d-11ea-8fc9-98a5f278157d.png)

RaymondSchnyder:
Ive spent 2 hours searching for the bug, but it was actually right there all the time.
Its a simple assignement bug, my merge request is right here:
https://github.com/EliasKotlyar/Xiaomi-Dafang-Hacks/compare/master...RaymondSchnyder:patch-1

Already tested on my local dafang, works fine.

dvybornov:
I've make the changes in telegram-bot-daemon.sh but it doesn't help.
Bot sends me alerts but don't respond on commands.

RaymondSchnyder:
> 
> 
[...]

PR #1484, opened 2020-06-08
Fix telegram bot not replying (#1470)

RaymondSchnyder:
Fixxes https://github.com/EliasKotlyar/Xiaomi-Dafang-Hacks/issues/1470.
Its an assignement issue:
on Line 96, the variable messageVal is checked if empty, and if yes, messageAttr is set to "edited_message".
Now on line 97, messageVal is always set to a json value, which may not be there. This line should actually be added to line 96 after it has been checked, it the messageVal is empty.
This fix ensures that messageVal is only set to the json value of '.result[0].edited_message' if messageVal was empty before.


jmtatsch:
Thank you for figuring that out.

PR #1489, opened 2020-06-11
Merging conf

gysmo38:
I re post my PR like #1471
I think again about the filter. I think it is not a good idea. I prefer to have a list of files that can be merge instead of a list of file that can not be merge.
The reason is, if someone add file that we can not merge there will be a problem with my script.
It easy to add a file name in the script or maybe we can have a file with the list of conf files that we can merge.


jmtatsch:
Maybe a whitelist is better after all

PR #1495, opened 2020-06-14
[UI] Add multi selection for delete in sdcard files

gysmo38:
Add multi selection for delete in sdcard files.
Bug fixies in index.html with datatable.min.css path
Move images for datable in his directory


Combie81:
I'm not seeing checkboxes to select multiple files? Have done a full refresh, am I missing something?

gysmo38:
There is no checkboxes but you can directly click on rows you want to select.
I will add buttons for select all/none in a new PR.

PR #1518, opened 2020-06-23
Adding a motion events graph for the images saved on SD card

andrempo:
image


jmtatsch:
Oh, 😮 I have been dreaming of that for a long time already. 🙏 you so much!

PR #1539, opened 2020-07-18
[Push To Talk] Added functionality to playback sound from the web

sergioamr:
Created a simple minimal UI that has a Push to Talk functionality. 
A basic send button that uploads the audio file and plays it back on the Dafang with a predefined volume.

Using standard javascript I create a WAV in 8KHZ that works with the audioplay binary.

It uses a temporal file at /tmp/playback.wav saved by a cgi called audio_upload.cgi

[Tested]
(PASS) Chrome Version 83.0.4103.116

(FAIL) Firefox 74.0 (64-bit) gives the following error while setting the sample rate to 8k:
    Connecting AudioNodes from AudioContexts with different sample-rate is currently not supported.

This could be fixed by supporting more formats at WAV level.

[Notes]
* The Audio apis require a secure HTTPS connection, therefore it will not run without certificates.

* I find an issue with the audio streaming, it seems that audioplay changes the bitrate of the audio and i have to restart the rtsp if I want to hear back correctly.

* File is not cropped at the end, so it has the end of the form header, I think the WAV format has a content size on the header and it is probably ignored. 

image

[Third Party Code]
It uses this library:
https://github.com/mattdiamond/Recorderjs

And the web code is based on:
https://blog.addpipe.com/using-recorder-js-to-capture-wav-audio-in-your-html5-web-site/

[Thanks!]

Thanks a lot for the work on this, I am using it a lot to monitor my cats and my 3d printing machines.

### DIFF
--- a/firmware_mod/www/cgi-bin/audio_upload.cgi
+++ b/firmware_mod/www/cgi-bin/audio_upload.cgi
@@ -1,0 +1,14 @@
+#!/bin/sh
+
+echo "Content-type: text/html"
+echo "Pragma: no-cache"
+echo "Cache-Control: max-age=0, no-store, no-cache"
+echo ""
+
+if [ "${REQUEST_METHOD}" = "POST" ]
+then
+    in_raw=`dd bs=1 count=${CONTENT_LENGTH} 1>/tmp/playback.wav`
+    sed -i -e '1,/Content-Type:/d' /tmp/playback.wav
+    echo " CONTENT LENGTH ${CONTENT_LENGTH}"
+    /system/sdcard/bin/audioplay /tmp/playback.wav 60
+fi
--- a/firmware_mod/www/play/audio_recordings.js
+++ b/firmware_mod/www/play/audio_recordings.js
@@ -1,0 +1,190 @@
+//webkitURL is deprecated but nevertheless
+URL = window.URL || window.webkitURL;
+
+var gumStream; //stream from getUserMedia()
+var rec; //Recorder.js object
+var input; //MediaStreamAudioSourceNode we'll be recording
+
+// shim for AudioContext when it's not avb.
+var AudioContext = window.AudioContext || window.webkitAudioContext;
+var audioContext //audio context to help us record
+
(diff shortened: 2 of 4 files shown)


sergioamr:
What it the process for PRs on this firmware? Best

jmtatsch:
Hi,

Sorry I totally forgot about this PR. 

For new functionality you should create the PR against the beta branch.

Also, can you please integrate this functionality into the main ui?
That way people will see/use it and it is not forgotten 😉

Then we will merge it.

sergioamr:
> Also, can you please integrate this functionality into the main ui?
> That way people will see/use it and it is not forgotten 😉

Sure!

jmtatsch:
Any progress on integrating this in the main ui?

nikopaulanne:
Please, continue on this project! This would be nice solution to 2-way audio replacement.

ortegafernando:
Hi, is it implemented?

BiatuAutMiahn:
I would love to see this implemented

sergioamr:
Sorry, I don't have a camera now, I damaged the ROM while trying to reflash the firmware trying to recover it.
The functionality works, I know it only needs to be added as suggested by EliasKotlyar to the main menu a integrated into the system. It is a small change but I don't own a working camera anymore.

Curtisdhi:
Something i've notice is the camera recording audio becomes "chipmunk" high pitch after using push to talk function. Any idea whats happening?

sergioamr:
@Curtisdhi It is related to how the audio has to be encoded for it to work with the device.

It is explained on the PR about how the system handles the bitrate of the playback. To be able to play, the audio has to be in a format the that installed program that plays audio can understand.

"I find an issue with the audio streaming, it seems that audioplay changes the bitrate of the audio and i have to restart the rtsp if I want to hear back correctly."

I used to just restart the streaming after talking, so the playback bitrate was back to the right format, but it takes a few seconds.
The real solution was to include another audio player that could support file formats with variable bitrates.

Sadly I don't have this camera anymore, so I never managed to get it finished, another unfinished project on my life.

Issue #1423, opened 2020-05-11
Missing reference frames


dewi-ny-je:
I'm recording video using RTSP H264 and codec "VBR 5000 kbps 15 fps".
Sometimes, especially when the motion in the frame is not extensive, I get the whole 13 seconds of recording (even if I set 10 seconds...) without a single reference frame.
This makes the video basically useless.

A reference frame should be introduced every second in the stream to ensure properly recorded videos.

fundef1:
I believe the script use avconv to concatenate the images into a video.
Ffmpeg is probably a better solution – if somebody could compile that…

I’ve use ffmpeg before to grab (MJPEG) streams directly from the RTSP stream.
(mjpeg because it’s way more efficient as it avoids an entire encode decode cycle)
But ffmpeg can also be use to concatenate images into a video.

nanoamp:
If you have motion-triggered recording from the RTSP server then (providing you have set `video_use_rtsp` to true) it'll buffer the raw RTSP data directly, rather than using avconv.

So, the fix would likely be upstream with the v4l2rtspserver, persuading it to insert more frequent I-frames. The config for the server is setup in _controlscripts/rtsp-h264_ using data from _config/rtspserver.conf_. Looking [upstream](https://github.com/mpromonet/v4l2rtspserver/issues/142) it's suggested that using the `-c` parameter for the rtspserver might help. Adding it to the `RTSPH264OPTS` in the conf file doesn't seem to make much difference on my cam though.

I'd be grateful if anyone else can come up with a way to improve this.

sshaikh:
I can confirm that -c did nothing for me - the GOP/iframe/keyframe seems set to 84 regardless of other settings. So yeah, probably best to lobby on the upstream issue.

sshaikh:
The author of upstream suggests another program, v4l2-ctl, can be used to control keyframes: https://github.com/mpromonet/v4l2rtspserver/issues/142#issuecomment-651394757

Is there a quick way to compile the binary for dafang?

nik0:
@sshaikh You have to know that  mpromonet/v4l2rtspserver is originaly for Raspberry. So the underlying API are different.

If someone want to play with the SDK, documentation:  https://github.com/tssva/Ingenic-SDK-T10T20-3.12.0-20180320/tree/master/doc
He is welcome to find to right API that controls the frequency of i-frame ...

sshaikh:
[...]

PR #1612, opened 2020-11-01
Merge beta and master branches

dewi-ny-je:
I merged the beta and master, since the beta has improvements which are working fine and which solve issues mentioned repetedly, such as the missing reference frames, which is major #1423 

The merge of beta and master showed some conflicts, which were however trivial to solve.

The merged branch has been tested on my camera and it seems to work. 

I am not an experienced coder, so I suggest carefully reviewing the proposed merge.


jmtatsch:
Ah great, you already did that. Will review it this week.

dewi-ny-je:
I just used GitHub Desktop to merge the two and I reviewed the conflicts... I didn't go deeper than that.
Also, except the redesign of the UI, if you check the commits in the two branches since spring, they don't look so incompatible.

Issue #1637, opened 2020-12-08
Differences between Web UI services and MQTT


benru:
I am interested in fixing some differences between the Web UI and MQTT interfaces to make them more equal. I am happy to do the pull request for these but wanted to discuss the issues first.

- MQTT still exposes separate "h264 rtsp server" and "mjpeg rtsp server" topics when the rest of the code has combined them into a single server. I propose to remove these topics and replace them with an "rtsp server" that behaves the same as the Web UI.
- MQTT has a topic "night mode" that allows you to manually enable/disable night mode on the camera. This is useful in that it sets the IR led/filter correctly but also switches the camera to black/white mode. I cannot find this feature on the Web UI, so I'd like to add it as part of the Camera Settings on the home screen.
- MQTT has the ability to turn on/off the recording feature. I think it may be useful to also have the ability to control the timelapse feature.

I just wanted to raise an issue before making the changes in case anyone else has any feedback about this area.

benru:
For reference here is my analysis of the features on the Web UI vs the current MQTT topics:

Webpage   - Home | Web page - Services | MQTT
-- | -- | --
Auto Night Detection | auto-night-detection | night mode auto
  |   | night mode
IR Led |   | ir led
IR Filter |   | ir filter
Blue LED |   | blue led
Yellow LED |   | yellow led
  | debug-on-osd |  
  | ftp_server |  
  | mqtt-control |  
  | mqtt-status |  
  | onvif-srvd |  
  | recording | recording
  | rtsp |  
  |   | h264 rtsp server
  |   | mjpeg rtsp server
  | sound-on-startup |  
  | telegram-bot |  
  | timelapse |  
Motion Detection |   | motion detection
Motion Send Emails |   | motion send mail
Motion Send Telegram |   | motion send telegram
Motion Blue LED |   |  
Motion Snapshots |   |  
Motion MQTT Message |   |  
Motion MQTT   Snapshot |  
Motion MQTT Video |   |  
  |   | motion tracking

jmtatsch:
a) agreed
b) agreed
c) agreed

You are welcome to make the naming (and the menu structure) more coherent if you like. 
I don't really like the naming on the services page. Home and mqtt namings are already quite aligned but surely can be perfected if given some thought.

benru:
If I'd been paying attention to the beta branch I would have seen that you've already done a) in 61a02eaec941b8f2193855b05709754252ce444c

[...]

PR #1640, opened 2020-12-08
Fix differences between Web UI services and MQTT

benru:
Fix differences between Web UI services and MQTT as described in #1637 

- adds a slider for manual Night Mode to the Web UI
- adds an MQTT topic to control the timelapse service
- changes to the descriptions of MQTT topics (not their paths!) to match the Web UI and have proper capitalisation
- added a list of friendly names for the services page that gives more user-friendly names to the current list of services

If a new service is added in the future the "friendly name" list should be updated, or the display name will default to the filename just as it did before.


jmtatsch:
Thank you very much. Improvements are very much appreciated 🙏

PR #1647, opened 2020-12-19
Fix select elements in safari

xdissent:
The visual selection doesn't reflect the actual value of the select box in safari.


jmtatsch:
Does that still work for firefox and chrome?
Can you please open the pull request against the beta branch?

xdissent:
It does and I will!

jmtatsch:
Thank you for your contributions.

PR #1648, opened 2020-12-19
Ignore motion events when night mode is switched on/off

xdissent:
This adds a configurable 30s period in which motion events are ignored when night mode is toggled. This prevents my cameras from sending a bunch of notifications at dusk and dawn while the night mode is settling.


jmtatsch:
Can the night_delay parameter be renamed to something more explaining like night_mode_event_delay?
And please also make the PR against beta.

jmtatsch:
Great addition by the way. It is quite annoying to get these false alarms.

PR #1649, opened 2020-12-19
Mdns responder

xdissent:
Adds (and autoruns by default) mDNSResponder and advertises the rtsp stream. This allows other hosts to discover both the hostname and the rtsp stream address via mDNS. (Now you can ping your camera at dafang.local)


jmtatsch:
Also against beta please and without the binary.
I will need to build that myself as we try not to check in compiled binaries.

jmtatsch:
I'll hopefully get around to build the mdns responder the next days and check it in.

PR #1650, opened 2020-12-19
GitHub token for updates

xdissent:
This PR adds a UI for a GitHub token which will then be used for checking for updates and while updating. (Basic user/password authentication has been deprecated by GitHub)


jmtatsch:
Looks good to me.
Can you also please do the PR against beta?

PR #1655, opened 2020-12-22
Network config web ui and fallback wifi hotspot

xdissent:
Adds web ui for configuring network settings and a fallback wifi hotspot to allow configuration of wifi without any previous config, or when wpa_supplicant fails to connect for 60 seconds (configurable). While in hotspot/access point mode, wifi is scanned every 30 seconds (configurable) and if the configured wifi SSID is found, wifi switches back on and the hotspot turns off. The default access point SSID is `dafang-<mac address>` (also configurable). In hotspot mode the web ui can be accessed at https://dafang.local (if mDNSResponder is on) or https://10.0.0.1

### DIFF
--- a/firmware_mod/config/hostapd.conf.dist
+++ b/firmware_mod/config/hostapd.conf.dist
@@ -1,0 +1,14 @@
+interface=wlan0
+ctrl_interface=/var/run/hostapd
+channel=6
+driver=nl80211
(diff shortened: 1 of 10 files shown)


jmtatsch:
What wonderful christmas 🎁 you gift to the Dafang community.

PR #1692, opened 2021-03-01
SSH management panel

timstanley1985:
Still a work in progress but i would like to add SSH configuration options to the front end.

- [x]  Change SSH port
- [x] Add authorized_key
- [x] Add option to disable password login

### DIFF
--- a/firmware_mod/www/cgi-bin/ui_system.cgi
+++ b/firmware_mod/www/cgi-bin/ui_system.cgi
@@ -25,6 +25,9 @@ if [ -n "$F_cmd" ]; then
     echo "scan_interval#:#$(get_config /system/sdcard/config/wifi.conf scan_interval)"
     echo "ap_ssid#:#$(get_config /system/sdcard/config/hostapd.conf ssid)"
     echo "usb_eth#:#$([ -f /system/sdcard/config/usb_eth_driver.conf ] && echo on || echo off)"
+    echo "ssh_key#:#$(cat /system/sdcard/config/authorized_keys)"
+    echo "ssh_port#:#$(get_config /system/sdcard/config/ssh.conf ssh_port)"
+    echo "ssh_password#:#$(get_config /system/sdcard/config/ssh.conf ssh_password)"
   ;;
   save_config)
 	if [ -n ${F_hostname} ]; then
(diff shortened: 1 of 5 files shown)


timstanley1985:
I am having problems with the ssh_key, it keeps substituting spaces with + symbols.  Ant thoughts on how I could get around this?

jmtatsch:
Can you point out on which line the problem with the spaces occur?

firmware_mod/www/system.html
```diff
@@ -66,6 +69,23 @@ <h1> System configuration</h1>
         <label for="usb_eth">Enable USB Ethernet (Disables Wifi)</label>
   </div>
 
+  <div id="ssh" class="w3-container tab" style="display:none">
+        <p></p>
+        <label>SSH Authorized Keys</label>
+        <input id="ssh_key" class="w3-input">
```
timstanley1985, line 75: If I enter text here, for example "my test key"

firmware_mod/www/cgi-bin/ui_system.cgi
```diff
@@ -78,6 +81,16 @@ if [ -n "$F_cmd" ]; then
 		echo "<p>Setting wifi password to: $wifi_password</p>"
 		wpa_config_set psk "\"$wifi_password\""
 	fi
+  if [ -n ${F_ssh_port} ]; then
+		ssh_port=$(printf '%b' "${F_ssh_port//%/\\x5}")
+		echo "<p>Changing SSH port to: $ssh_port</p>"
+		rewrite_config /system/sdcard/config/ssh.conf ssh_port "$ssh_port"
+	fi
+  if [ -n ${F_ssh_key} ]; then
+		ssh_key=$(printf '%b' "${F_ssh_key//%/\\x}")
```
timstanley1985, line 90: The spaces in the key get swapped for +'s for example "my+test+key"

firmware_mod/www/cgi-bin/ui_system.cgi
```diff
@@ -78,6 +81,16 @@ if [ -n "$F_cmd" ]; then
 		echo "<p>Setting wifi password to: $wifi_password</p>"
 		wpa_config_set psk "\"$wifi_password\""
 	fi
+  if [ -n ${F_ssh_port} ]; then
+		ssh_port=$(printf '%b' "${F_ssh_port//%/\\x5}")
```
fhl206, line 85:
```suggestion
		ssh_port=$(printf '%b' "${F_ssh_port//%/\\x}")
```

timstanley1985: Yes, will fix

firmware_mod/www/cgi-bin/ui_system.cgi
jmtatsch: Plus signs in CGI parameters mean spaces. Apparently that's in the CGI spec.
Maybe run javascript to encode the spaces to %20 before sending to the key to the cgi script.
Then in the shell script use regex to replace to decode the spaces to real spaces again.

timstanley1985: Thanks, that worked perfectly.  Will submit a final version of this patch tomorrow

timstanley1985:
Thanks for your help, I think this is ready to merge unless you have any other comments

PR #1701, opened 2021-03-08
Stopping RTSP will also Stop Its Dependent Services Properly

### LINKED ISSUE #1686: Restarting RTSP should restart recording too (and maybe other script/service dependency problem)
I notice that there is no fancy things such as `systemd` or alike in this tiny microcontroller :)

If the RTSP is restarted, the recording simply crash, you can reproduce this through the web UI or through the controller scripts.

There maybe other script/service dependencies that are still not properly handled.

Maybe the quickfix is to watch the `/run/*.pid` to see if there is inconsistency, i.e. if the PID is listed in `/run` but not actually running, the script must be started again.

I can contribute if I have time, but you guys are welcome if you can do it first!

yerlandinata:
Without this, the dependent services will just be killed, but the .pid will still remains, making restarting from web UI impossible.

With this changes, when the rtsp is stopped, the dependent services will also be stopped properly. Therefore, the web UI can be used for restarting the dependents.

resolves #1686

### DIFF
--- a/firmware_mod/controlscripts/rtsp
+++ b/firmware_mod/controlscripts/rtsp
@@ -20,6 +20,7 @@ fi
 if [ -z "${CODEC+x}" ]; then
     CODEC="H264"
 fi
+
 status()
 {
   pid="$(cat "$PIDFILE" 2>/dev/null)"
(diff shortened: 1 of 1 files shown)


yerlandinata:
If there is a way to bring in something like `systemctl` or `services` that'd be great.
But for now, with this PR, one bug is fixed.

PR #1720, opened 2021-04-13
Fix: waiting for all background jobs to finish before send snapshot

Nazim-ua:
Fixed deletion of a temporary file before the snapshot is complete


timstanley1985:
Looks good to me! 😃

PR #1724, opened 2021-04-26
Fix missing directory structure on letsencrypt

TheRealWaldo:
When creating the cron script, the directories may/may not already exist.  In my case, they did not, so the script fails.

This makes sure the directory exists first to avoid the error.

### DIFF
--- a/firmware_mod/config/ssl/letsencrypt_setup.sh
+++ b/firmware_mod/config/ssl/letsencrypt_setup.sh
@@ -53,6 +53,7 @@ fi
 
 
 ## Adding cronjob to keep the cert updated
+mkdir -p ${CONFIGPATH}/cron/periodic/weekly/letsencrypt
 cat > ${CONFIGPATH}/cron/periodic/weekly/letsencrypt <<EOF
 #!/bin/sh
 PATH=/system/sdcard/bin:/system/bin:/bin:/sbin:/usr/bin:/usr/sbin


Dopeyr:
@jmtatsch @TheRealWaldo Doesn't that create a directory called 'letsencrypt'? AFAIK that should be a file?

jmtatsch:
It creates a directory that might very well be missing. Seemed fine to me.

Dopeyr:
Hi, the issue seems to be that it creates a directory called 'letsencrypt' then tries to cat a file to it:

```
a@ubuntu-server-docker:/tmp$ mkdir -p cron/periodic/weekly/letsencrypt
a@ubuntu-server-docker:/tmp$ cat > cron/periodic/weekly/letsencrypt <<EOF
> !/bin/sh
> PATH=/system/sdcard/bin:/system/bin:/bin:/sbin:/usr/bin:/usr/sbin
> EOF
-bash: cron/periodic/weekly/letsencrypt: Is a directory
```

So I end up with:

`-bash: cron/periodic/weekly/letsencrypt: Is a directory`

I've not run this on the device though. In case sh works differently there, I could be wrong.

jmtatsch:
I guess you are right. Path should only be created up to weekly. Then we can cat the job there.

TheRealWaldo:
Yeah, that's a nistake on my part.  I had recreated the commit a couple times and copy/pasted blindly the last time without looking.

Sorry guys, it should be without the letsencrypt.

Fixed in #1725

jmtatsch:
Thank you for fixing things and @Dopeyr thank you for keeping an eye on things 👍🏻

PR #1726, opened 2021-04-27
Root CA updates and more cron fixes

TheRealWaldo:
At some point, the root crontab was added to the git source, but the original script that added them was not updated to reflect.

Also, updates the root ca file and the updater to pull from the new location.


jmtatsch:
Is this working on the device?

TheRealWaldo:
Working on my WyzeCam v2.  Don't have other models to test.

Issue #1667, opened 2021-01-16
New Wyze V2 cannot load CFW


pbruen:
Looks like WYZE V2 cameras that have had newer firmware loaded on them do not allow CFW to install properly.  Doing so returns:
solid orange light for ~3s. a click, light blinks out briefly after another 3s, and repeat.  
Loading an older official firmware prior to CFW  does not improve things

If there is a method around this, I'd love to hear it, as these devices are worthless to me with firmware that requires external connectivity (all the official firmware).  I've seen a few other people describe this in a few different places, but no one seems to respond and (at least in git) the issue gets closed as stale.  :(

ardlc0:
Can confirm that this issue is also happening on my end. I have a Wyze cam V2 purchased from Amazon in December. I have the exact same issues as @pbruen described. Once I write the cfw-1.1.1bin file onto the micro sd card, the flashing process looks like it begins because the light turns from orange to blue, but then it goes back to orange for about 3-6 seconds then turns off, turns back on for another 3 seconds, makes a clicking noise and then turns back off. This repeats until I unplug the device from USB.

Steps in order to replicate issue:
1. **Mount micro sd to filesystem and erase all contents on micro sd.**
     
     Note: I have tried multiple types of micro sd cards and formatted them completely to erase all contents and old partitions. First was a 256mb Kingmax, the second was a 4gb Sandisk, and the third was a 32gb Sandisk ImageMate. For the 4 gb and 32 gb micro sd cards I created the recommended 512mb primary partition with boot flag on both. I also tried using the full sd card capacity for 4gb and 32gb and still no luck. I used the fat32 format for all sd cards tested. I also tried with gpt and ms-dos partition tables. No luck with those either.

2. **Download cfw bin file for the Wyzecam V2 file**. Located here -> https://github.com/EliasKotlyar/Xiaomi-Dafang-Hacks/raw/master/hacks/cfw/wyzecam_v2/cfw-1.1.bin
     
     Note: I tried also the T2 and Xia0mi dafang cfw bin files. Not able to flash using these either. 

3. **Copy bin file to the micro sd card root. Rename cfw-1.1.bin to "demo.bin"**
4.  **Safely unmount the micro sd and insert into Wyzecam V2 while camera is off and not connected to USB.**
5. **Press the setup button and insert the usb cable to supply power to camera.**
6. **Wait ~20 seconds until orange light turns blue. Release the setup button after light has turned blue.**
[...]

PR #1729, opened 2021-05-06
Updated cfw for newest wyzev2

McNutnut:
- Added latest Wyze v2 RTSP firmware source
- Synced hacked rc file latest Wyze v2 RTSP firmware
- Matched blocksize for wyzev2 to match latest RTSP firmware
- Added updated cfw for latest Wyze v2 cameras (FCCID ending in C02)

Resolves https://github.com/EliasKotlyar/Xiaomi-Dafang-Hacks/issues/1667 and https://github.com/EliasKotlyar/Xiaomi-Dafang-Hacks/issues/1702

Tested this across 3 Wyze cams no issues.   Users also reported success with the updated firmware in the above issue threads


Real-Time-Kodi:
Confirmed this fixed my problem with my wyze cam v2 FCC ID: 2ANJHWYZEC02

jmtatsch:
Looks good to me 👍. Thank you for contributing.

lacojim:
This does not seem to be compatible with the open source U-Boot bootloader.

lacojim:
I used the same one I used for the other three cameras. dafang_128mb_v2.bin. It rendered the camera unbootable. I only had to use the cfw1.2.bin on this one camera. The uboot worked fine with the others using cfw1.1.bin

Real-Time-Kodi:
> Do you get an error message? What uenv do you use?
Check my open issue on the uboot repository
https://github.com/Dafang-Hacks/uboot/issues/18
Camera loads into uboot but cannot seem to access the flash memory. I can load cfw 1.2 manually and it works fine, I suspect the problem is with uboot and this specific revision of the camera, not with the cfw itself.

jmtatsch:
Probably the offset where the boot partition is mounted in the new cfw has changed so you would need put that new address in your uEnv file.

lacojim:
FYI, I have two cameras with FCCID 2ANJHWYZEC02. One worked with the cfw1.1 and the other needed cfw1.2. There is probably only a limited amount of these in distrubution, but there are definitely two different cams with same FCCID.

PR #1743, opened 2021-06-08
Add Push To Talk

StevenJonSmith:
Taken the previous work of sergioamr and modified to be compatible with the current UI. Tested working on a UK Neos Smartcam.

Takes a voice recording on button hold and then sends to camera for playback on release.


jmtatsch:
Very nice, thank you 👍🏻

PR #1748, opened 2021-06-22
Change absolute URL's to relative

c---:
There were a few places in the code that had absolute paths for the URL. This
causes problems if you want to put the camera behind a reverse proxy or similar.
After these changes I'm able to run my cameras behind nginx so I can access all
of them remotely through one web site.

Hopefully I found all places that needed to be fixed. Sometimes a new one pops
up. I did my best to go through every page in the interface while watching the
web console for errors.


c---:
This shouldn't affect normal operation in any way, BTW.

PR #1754, opened 2021-07-04
Implement Recording Rollover

yerlandinata:
For me recording is important, because the motion detection is somewhat unreliable.
Having one large file is problematic, not to mention it will keep recording to that file forever.

With this pull request, the recording output file will be changed once every 30 minutes.

- implemented using cron
- roll every 30 minutes, not configurable via web (crontab)
- on/off toggle is configurable via web UI
- retention duration is configurable via web UI

TL;DR implementation: simply restarting the recording service once every 30 minutes, because actually splitting the file maybe impossible for some devices (mine only have 1MB free memory)

### DIFF
--- a/firmware_mod/scripts/roll-recording
+++ b/firmware_mod/scripts/roll-recording
@@ -1,0 +1,28 @@
+#!/bin/sh
+
+roll()
+{
+	if [ ! -f /system/sdcard/config/recording.conf ]; then
+		cp /system/sdcard/config/recording.conf.dist /system/sdcard/config/recording.conf
+	fi
+
+	if [ ! -f /run/recording.pid ]; then
+		exit
+	fi
+
+	. /system/sdcard/config/recording.conf
+	if [ "$ROLL_RECORDING_ENABLE" = true ]; then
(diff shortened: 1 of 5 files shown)


Kaan88:
This stale bot...

I found this usefull and applied it to my installation. But I think there is a bug, check for "/run/recording.pid" always returns true and the recording never starts. So I had to comment it out.

Other than that the only issue is that this feature would only stop when the cron job executes. So If I make it roll every two hours, it might not stop recording for a long time. Could this be implemented as a service that adds the cron job only then?

Kaan88:
$DCIM_PATH is undefined, so the retention doesn't work as intended.

firmware_mod/scripts/roll-recording
```diff
@@ -0,0 +1,28 @@
+#!/bin/sh
+
```
dewi-ny-je, line 2: missing
. /system/sdcard/config/rtspserver.conf
to load $DCIM_PATH

dewi-ny-je (commented):
The idea is good, corrected a minor thing.

Issue #1767, opened 2021-07-30
Telegram video alerts are no longer sent


nanoamp:
Neos SmartCam running latest beta commit (21f00d7).

### Description

Since I updated to a version after #1709 was merged, I've not had any video alerts via Telegram.

### What did you do to debug the issue

I replicated the new behaviour of _detectionOn.sh_ by creating a test video:
`openRTSP -4 -w 1920 -h 1080 -f 20 -d 10 -b 1500000 rtsp://REDACTED:REDACTED@127.0.0.1:8554/unicast > /tmp/test.mp4`

and then trying to strip the audio:
`avconv -i /tmp/test.mp4 -c:v copy -c:a copy /tmp/test-telegram.mp4`

This fails with the following error:
```
avconv version v13_dev0-1541-ge368b0cbf, Copyright (c) 2000-2018 the Libav developers
  built on May 17 2018 20:39:55 with gcc 4.7.2 (Ingenic r2.3.3 2016.12)
[mov,mp4,m4a,3gp,3g2,mj2 @ 0x1776040] invalid size 8 in stsd
[mov,mp4,m4a,3gp,3g2,mj2 @ 0x1776040] error reading header: -1052488119
test.mp4: Invalid data found when processing input
```

This is code which appears to have worked for @timstanley1985 in #1658, so I'm not sure why my instance would create an MP4 which isn't handled by avconv.

nanoamp:
Output from openRTSP:
```
Created new TCP socket 3 for connection
Connecting to 127.0.0.1, port 8554 on socket 3...
...remote connection opened
Sending request: OPTIONS rtsp://REDACTED:REDACTED@127.0.0.1:8554/unicast RTSP/1.0
CSeq: 2
User-Agent: openRTSP (LIVE555 Streaming Media v2020.05.15)


Received 152 new bytes of response data.
Received a complete OPTIONS response:
RTSP/1.0 200 OK
CSeq: 2
Date: Fri, Jul 30 2021 21:29:43 GMT
Public: OPTIONS, DESCRIBE, SETUP, TEARDOWN, PLAY, PAUSE, GET_PARAMETER, SET_PARAMETER

Sending request: DESCRIBE rtsp://REDACTED:REDACTED@127.0.0.1:8554/unicast RTSP/1.0
CSeq: 3
User-Agent: openRTSP (LIVE555 Streaming Media v2020.05.15)
Accept: application/sdp

[...]

PR #1769, opened 2021-08-02
More robust Telegram video handling

nanoamp:
This PR is to prevent unparseable audio formats from blocking video dispatch to Telegram. Also provides a bit more debug/log info, and brings feature parity to the bot /command with the camera UI.


nanoamp:
Discussion in #1767

PR #1789, opened 2021-12-03
Fix "FTP only record mjpeg" problem

xuan-w:
FTP video recording didn't use record_video()'s output.
So no h264 rtsp was available, only 1fps mjpeg video.

Now remove the duplicated FTP mjpeg code, so that FTP upload can
be controlled by the same variables as other upload methods.

1. Add ftp_video test for triggering record_video()
2. modify ftp video upload so that it use the video_tempfile
generated by record_video()

### DIFF
--- a/firmware_mod/scripts/detectionOn.sh
+++ b/firmware_mod/scripts/detectionOn.sh
@@ -214,6 +214,7 @@ send_snapshot &
 # Then, record video (if necessary)
 if [ "$save_video" = true ] ||
    [ "$smb_video" = true ] ||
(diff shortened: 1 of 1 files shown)


jmtatsch:
Hi, 

thanks of contributing.
The code you removed was quite interesting in that it writes frame by frame to a ftp server.
In case a criminal comes by, him unplugging the camera is pointless because its already on the off premise ftp server. So this is why I really want to keep the old path around.

If you can keep both code paths alive somehow I will happily merge your code.

Issue #1797, opened 2022-01-22
setting the correct date / time


jammmyb:
i am running this on two different cameras.

one is a Neos SmartCam and the other is the Xiaomi Dafang.

the Dafang has the correct date / time, but the smartcam is Current time is **Thu Sep 17 02:26:14** **BST 1970** - GMT0BST,M3.5.0/1,M10.5.0

![dafang](https://user-images.githubusercontent.com/48286264/150637766-d56ae074-99d5-4103-8d9c-e284c6ffb267.png)
![smartcam](https://user-images.githubusercontent.com/48286264/150637769-49b95b27-c291-4f5c-a526-fd0e95dc336a.png)

how do i fix this please?

mercs600:
I noticed the same issue, any solution ?

c---:
If you ssh in to the camera then run:
```
ntpd -d -w -p time.google.com
```
Do you get replies?

kittenarmy:
> ntpd -d -w -p time.google.com

I get response `ntpd: bad address 'time.google.com'`

c---:
@kittenarmy My ntpd command was just to test the network. If you want to set the time then remove the `-w` option.

kittenarmy:
> @kittenarmy My ntpd command was just to test the network. If you want to set the time then remove the `-w` option.

Ahh thank you!

jammmyb:
![ssh](https://user-images.githubusercontent.com/48286264/164041900-6e4901e9-943c-47d1-94c6-a47bf08d85bb.png)

still wrong date/time

what can i try next?

c---:
If you want to debug _and_ set the time then you need to use something like:
```shell
ntpd -d -n -p time.google.com
```
Without the `-n` or `-w` option it's going to fork in to the background and won't output the debugging messages.

It might be a good idea to make sure there are not multiple ntpd running. `ps|grep ntpd`

In theory ntpd is started from /system/sdcard/run.sh and can be verified that it's running by using the same ps command `ps|grep ntpd`. If the debugging tests succeed and the automated service is active and running but the time is not being set then I'm not sure what could be going on. Possibly the ntp server in ntp_srv.conf is incorrect or for some unknown reason the background service can't set the system time.

kittenarmy:
I ran `ps|grep ntpd` and got two lines:
```
  279 root      0:00 /system/sdcard/bin/busybox ntpd -p time.google.com
[...]

PR #1811, opened 2022-04-19
Update run.sh

kittenarmy:
Date time fix for Static IP #1797

### DIFF
--- a/firmware_mod/run.sh
+++ b/firmware_mod/run.sh
@@ -146,6 +146,13 @@ ntp_srv="$(cat "$CONFIGPATH/ntp_srv.conf")"
 timeout 30 sh -c "until ping -c1 \"$ntp_srv\" &>/dev/null; do sleep 3; done";
 /system/sdcard/bin/busybox ntpd -p "$ntp_srv"
 
+## Date time fix for static IP:
+if [ -f $CONFIGPATH/staticip.conf -a -f $CONFIGPATH/resolv.conf -a -f $CONFIGPATH/defaultgw.conf ]; then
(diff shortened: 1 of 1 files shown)


jmtatsch:
@kittenarmy I dont understand why this would be needed. Why does ntp not work for you?

dewi-ny-je:
I also don't understand why a static IP would require different treatment of NTP, which is a higher level service

kittenarmy:
> @kittenarmy I dont understand why this would be needed. Why does ntp not work for you?

I've no idea why it doesn't work. Read what I tried in #1797. I can only get NTP to work over SSH.

firmware_mod/run.sh
```diff
@@ -146,6 +146,13 @@ ntp_srv="$(cat "$CONFIGPATH/ntp_srv.conf")"
 timeout 30 sh -c "until ping -c1 \"$ntp_srv\" &>/dev/null; do sleep 3; done";
 /system/sdcard/bin/busybox ntpd -p "$ntp_srv"
 
+## Date time fix for static IP:
+if [ -f $CONFIGPATH/staticip.conf -a -f $CONFIGPATH/resolv.conf -a -f $CONFIGPATH/defaultgw.conf ]; then
+  str=$(/system/sdcard/bin/curl -s --head http://google.com | grep ^Date: | sed 's/Date: //g')
+  d=$(date -d "$str" -D "%a, %d %b %Y %T %Z" +'%Y-%m-%d %H:%M:%S')
+  date -s "$d"
```
sharmay, line 153:
```suggestion
  date -u -s "$d"
```
Date/Time will be wrong, date/time retrieved via curl are in GMT (during my testing).

PR #1820, opened 2022-06-05
Add proper firmware image for Neos Smartcam

bengris32:
* Create a modified firmware image based on the original neos firmware so people can boot back to the Neos firmware instead of booting to the Wyze firmware.


jmtatsch:
Thank you for contributing, sorry it took so long to merge.

PR #1846, opened 2023-04-17
Prevent HTTPS redirects for 127.0.0.1

unqueued:
This is helpful for when a host is connected over an ssh tunnel, and sends "localhost" for the HOST header. If it is an ssh tunnel, it is already secured. And browsers have trouble with redirects to https://localhost

That means that a user can set up an ssh `ssh -L 8080:localhost:80 dafang.local, and then view the camera with http://localhost:8080`

Currently, it will try try to redirect the client to https://localhost:8080, which fails.

I came up with this when I had to use ssh jump hosts to connect to specific cameras behind a bastion server. And I think it is safe to assume that localhost does not require encryption.


unqueued:
Hey, I just noticed I had my PR wrong before. I've now based it off of the [beta](EliasKotlyar/Xiaomi-Dafang-Hacks/tree/beta) branch.

PR #1848, opened 2023-04-29
Multiple fixes & improvements related to volume and audio.

dbuezas:
## Fix setting HW Volume from web UI
* The input field had no ID, and the js script ignored that field

## Fix setting filter from web UI
* The path to setconf was wrong in ui_camera.cgi

## Expose SW volume in web UI
* This is useful for low volume cameras.
* Mine for example has very low sensitivity without SW volume turned up
image

## LIVE Update audio volume and filters, FPS and OSD settings from the Web UI
* Volume updates as soon as you release the mouse while scrolling the slider
* Without interrupting the stream
* It is not persisted unless one clicks save
* Useful for testing filters and volumes before saving

## Volumes (HW and SW) exposed via MQTT 
* Persisted
* Both scaled to appear as 0 to 100%
image

### DIFF
--- a/firmware_mod/scripts/common_functions.sh
+++ b/firmware_mod/scripts/common_functions.sh
@@ -881,3 +881,31 @@ wpa_config_get() {
 		grep "^[[:space:]]*$key=" "$wpa_config" | cut -d "=" -f2
 	fi
 }
+
+# Control the hw volume
+hwvolume() {
+	if [[ "$1" == "status" ]]; then
+		# Get the current volume value
+		local raw=$(/system/sdcard/bin/setconf -g h)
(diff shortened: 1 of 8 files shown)


dbuezas:
It works well, the only issue I am having is:
* Something is resetting the HW and SW volumes to 70 and -1 on each restart

And the only feature I'd like to add is:
* Expose them via mqtt

jmtatsch:

dbuezas:
Gladly, I wanted to understand the internals of this and ended up learning lots more about bash.

jmtatsch:
Haha yes, there is a reason why the repo is called defang hacks 😉

PR #1865, opened 2023-07-22
Improve install_sdcard.md documentation

jcplaza:
After reading the manual provided in install_sdcard.md manual i tried to follow it, but i couldn't boot mi camera with the wireless up, and it's difficult to know if the cam is up with the status led only

Inventigating I realized that have to uncomment some lines in rcS init script, so i tried to make a better manual to reflect it.

### DIFF
--- a/hacks/install_sdcard.md
+++ b/hacks/install_sdcard.md
@@ -1,19 +1,80 @@
+
 # Installation of the Open Source Camera Firmware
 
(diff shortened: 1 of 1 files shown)


jmtatsch:
definitely much improved. thanks

zanymuzik:
Thanks a lot for adding this information - it definitely more readable and much easier to follow now.
I was able to boot my dafang to sdcard based on this.
However, it is stuck on v0.0.1 and i am unable to update the UI at all.

So, the bigger question - is the sdcard method still relevant? and if yes, what is the way to update the firmware as the update option in the UI is a no-op.

jmtatsch:
Not really relevant at all to the average user. Interesting if one really wanted to run a completely own os.

zanymuzik:
I agree with you that this is more advanced than an average use case.
However, were you able to run a newer version of the UI (> v0.0.1) with this?

PR #1873, opened 2023-08-07
Fix 2023.8 repair warning

pplucky:
Fix 2023.8 repair warning 'Discovered entities with a name that starts with the device name'

This stops working in version 2024.2.0. Please address before upgrading.

Some MQTT entities have an entity name that starts with the device name. This is not expected. To avoid a duplicate name the device name prefix is stripped of the entity name as a work-a-round. Please inform the maintainer of the software application that supplies the affected entities to fix this issue.


jmtatsch:
Isn't that already fixed in beta?

pplucky:
> Isn't that already fixed in beta?

Missed it, sorry.

